PR #34, opened 2020-10-04
Dynamically populate language list on index page

### LINKED ISSUE #16: Have welcome/home page auto-populate concepts/languages
Right now they're static but as the thesauruses get split out from the app, they'll need to be more dynamic

colindean:
Fixes #16


geekygirlsarah:
I'm giving you the side eye because you added the enhancement (so thank you very much!) but also threw in some other things not in the issue and it's going to mean a few awkward changes later...  but this is good and I appreciate it and I'll work around the Scala additions.

So 👍 from me!

PR #41, opened 2020-10-05
Add unit tests to Django

### LINKED ISSUE #23: Feature: Add unit tests for Django
**Description:**
This project has kind of been thrown together over past Hacktoberfests and on Sarah's free time. It's not really had tests added to it because of it's sort of hacky nature. I'd like to have unit test get added in to ensure old features continue to work and new features won't break the site when they go live.

**Requirements:**
* Add some tests for the views
* Add some tests for the models (if they've been added at this point)
* Add some tests for the controllers (if they've been added at this point)

**Claiming This Issue:**
* Please leave a comment below indicating your interest. I can then assign it to you.
* If I don't see an update (comment, code, or PR), then I'll take your name off.
* You would always be welcome to reclaim it again.
* If you can't work on it, either leave a comment or remove yourself from the issue.
* If you have questions, please let me know!

Arodriguez340:
Hi, I added some test cases, and name spacing for templates. I still working on some views test.


web/templates/web/about.html
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html' %}
+{% extends 'web/base.html' %}
```
geekygirlsarah: Why did you add the `web/` back into these and move the files into a `web` directory?

web/urls.py
```diff
@@ -2,6 +2,8 @@
 
 from . import views
 
+app_name = 'web'
```
geekygirlsarah: Ok, I see you added this (therefore making the need to add "web" to assorted other areas. I'm curious why you added it as I was kind of thinking it wasn't needed for a single-app project. Is it more standard practice or am I breaking things by not having it?

geekygirlsarah:
I like where you're going with this. So far it has a 👍 from me whenever you feel it's ready to be done. We may need to look at the merge conflicts that will happen too since some files have changed since you moved them.

web/urls.py
Arodriguez340: It just a standard practice.

web/templates/web/about.html
Arodriguez340: When I was running some test to make sure that  the view was actually using the correct template, I ran into some errors. The test didn't match any if I just wrote the template name. therefore, I google it, and find out that the best approach was to add the namespacing, and then add the template path to the assertTemplateUsed(some_response, template path).

geekygirlsarah:
Hmm... this wasn't supposed to close yet, looks like deleting `develop` did that.

@Arodriguez340 I merged the `develop` branch into `master` so this may need to be edited or closed and reopened to continue work. Sorry about that. 

Also were you still working on this?

Issue #27, opened 2020-09-28
Feature: Convert non-existent language items from `null` code to empty object


geekygirlsarah:
Inside of a language item, code was `null` to symbolize it didn't exist in the language. This could just be an empty object (i.e. `integer_32_bit: {}`). 

Update JSONs and code for this.

colindean:
I _think_ this already works? https://github.com/codethesaurus/codethesaur.us/pull/34/files#diff-0390e01eb8c66734c21151cb8f4dae7dR38 seems to work?

vaibhvshrma:
Hey @geekygirlsarah, 
Wonderful idea behind the project. Could you assign the issue to me? I will make the required changes.
Thanks!

geekygirlsarah:
@vaibhvshrma I have already started working on it but haven't pushed changes yet, mostly because my changes fix 3 issues at the same time (which is why I went ahead and grabbed them myself). I appreciate the enthusiasm though, thanks!

Issue #26, opened 2020-09-28
Add loop for categories on compare/reference views


geekygirlsarah:
I took out the categories loop while on the Symfony->Django conversion. I should add that back in.

Issue #25, opened 2020-09-28
Remove nested `id` from thesauruses


geekygirlsarah:
A sub-concept in the data files has a `id` field inside of the object that also has an `id`. This can be removed.

PR #43, opened 2020-10-05
Fix #42 keyerror on /compare and make view more robust

### LINKED ISSUE #42: /compare/ throws a KeyError in develop branch
The /compare/ end point throws a key error. This is reproducible for multiple different language choices, few examples I tried out was Java and C++, C++ and C#.

vaibhvshrma:
Fixes #42 
Use intersection to find common topics and make view more robust

### DIFF
--- a/web/views.py
+++ b/web/views.py
@@ -1,4 +1,7 @@
-from django.http import HttpResponse, Http404, HttpResponseRedirect
+from django.http import (
+    HttpResponse, Http404, HttpResponseRedirect,
+    HttpResponseBadRequest,
+)
 from django.shortcuts import render, get_object_or_404
 from django.urls import reverse
 
@@ -39,8 +42,12 @@ def compare(request):
 
     lang1 = request.GET.get('lang1', '')
     lang2 = request.GET.get('lang2', '')
-    lang1_directory = meta_data_langs[lang1]
-    lang2_directory = meta_data_langs[lang2]
+    lang1_directory = meta_data_langs.get(lang1)
+    lang2_directory = meta_data_langs.get(lang2)
+
+    if not (lang1_directory and lang2_directory):
+        # TODO: Add meaninglful response
+        return HttpResponseBadRequest("Lang does not exist")
 
     concept = request.GET.get('concept', '')
 
--- a/web/views.py
+++ b/web/views.py
@@ -46,7 +46,6 @@ def compare(request):
     lang2_directory = meta_data_langs.get(lang2)
 
     if not (lang1_directory and lang2_directory):
-        # TODO: Add meaninglful response
         return HttpResponseBadRequest("Lang does not exist")
 
     concept = request.GET.get('concept', '')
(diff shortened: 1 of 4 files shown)


web/views.py
```diff
@@ -38,33 +42,46 @@ def compare(request):
...
+    lang1_directory = meta_data_langs.get(lang1)
+    lang2_directory = meta_data_langs.get(lang2)
 
-    with open("web/thesauruses/" + lang1_directory + "/" + concept + ".json", 'r') as lang1_file:
-        data = lang1_file.read()
-    # parse file
-    lang1_file_json = json.loads(data)
-    lang1_concept = lang1_file_json[concept]
-    lang1_friendlyname = lang1_file_json["meta"]["language_name"]
+    if not (lang1_directory and lang2_directory):
```
geekygirlsarah, line 48: This was an error check I wanted to add but hadn't gotten that far. Thanks!

web/views.py
```diff
@@ -38,33 +42,46 @@ def compare(request):
...
+    concept = request.GET.get('concept', '')
 
-    common_concepts =[]
-    for key in list(set.union(set(lang1_concept.keys()), set(lang2_concept.keys()))):
+    lang1_file_path = os.path.join(
+        "web", "thesauruses", lang1_directory, concept) + ".json"
+    lang2_file_path = os.path.join(
+        "web", "thesauruses", lang2_directory, concept) + ".json"
+
+    try:
```
geekygirlsarah, line 58: Thanks for adding this `try` block too. It was another thing I hadn't converted over from the old framework!

web/views.py
```diff
@@ -38,33 +42,46 @@ def compare(request):
...
+            data = lang2_file.read()
+            # parse file
+            lang2_file_json = json.loads(data)
+            lang2_concept = lang2_file_json[concept]
+            lang2_friendlyname = lang2_file_json["meta"]["language_name"]
+    except FileNotFoundError as fe:
+        return Http404
+
+    common_concepts = []
+    for key in (set(lang1_concept.keys()) & set(lang2_concept.keys())):
```
geekygirlsarah: I may need to test this. I need the equivalent of a union/full outer join on this and I can't tell if this will do the same thing. I need to match up language concepts if they both exist, but if one or the other doesn't exist, I still need it to show up and the other can be `null` or something. So I'll need to test this before approving.

vaibhvshrma: @geekygirlsarah I will save you some time, this does not work that way xD, I thought you intended only to show the intersection i.e. the common concepts only. The union was throwing a key error, I will make the required changes here. I will make it show up as NA if it does not exist in one, hope that will be fine. Let me know about anything else that you would like to show instead.

vaibhvshrma: @geekygirlsarah I have updated it to show an unsupported message as follows: "Unsupported or Not Needed". Felt this was more apt.

geekygirlsarah:
I appreciate the modifications and what you did with this. Thank you! I'll merge it now.

vaibhvshrma:
Thank you for the hacktoberfest accepted label, @geekygirlsarah! It was a pleasure to contribute. Please do let me know if there are any other issues/features to work on.

geekygirlsarah:
When I finish issues #25, #26, and #27 (which I've started and not finished, and they're all similarly tied together so I'm doing them at the same time), this should lay some more foundation for me to expand out the "thesaurus" parts of the site, and I should be able to start making _**TONS**_ of issues to help build out the actual usability of the site. I'll also want to collect stats on things but that involves some work on setting up a database instance and I'll have to do that.

PR #45, opened 2020-10-05
Issue #44: Unsupported items shouldn't be formatted as code.

### LINKED ISSUE #44: Feature: Items in a language that aren't supported shouldn't show up as if it was code
If a language item that doesn't exist in the language currently shows up as if it was code. This shouldn't show up as monospaced code items but as regular text, or be distinguished somehow differently.

image

nukefusion:

- [ ] Refactor
- [X ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## 📖 Description

This fixes #44. Also refactored out some repetitive code using template tags.

### DIFF
--- a/web/templates/compare.html
+++ b/web/templates/compare.html
@@ -1,4 +1,5 @@
 {% extends 'base.html' %}
+{% load templatetags %}
 
 {% block content %}
         <div class="container">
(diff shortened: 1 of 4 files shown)


geekygirlsarah (commented):
This does fix the issue I was wanting fixed, but it's a extra refactor of moving it to a separate file I wasn't exactly expecting. If you would make the same refactor to the `reference.html` template, I'll merge it in.

nukefusion:
Once the additional conditional logic is introduced into the template, duplicating it via copy/pasting starts to get ugly/messy/error prone and would be considered by many to be bad practice, so my feeling is that the refactoring was a necessary and natural part of this work in the interests of keeping it DRY. Agree that reference.html should be updated to use the inclusion as well, I will try to update later.

geekygirlsarah:
Oh I agree, I'm all for refactoring, but also the refactor didn't cover everything it should refactor and didn't know that it'd actually get duplicated much more. So if you do `reference` too I'd be cool with it!

geekygirlsarah (approved):
I think that final change will make this work.

Thanks for working on this!

PR #76, opened 2020-10-19
Resolve query string injection issues with sanitation

### LINKED ISSUE #74: Bug: Can inject bad things into query strings on Compare/Reference page
# Expected Behavior
Only languages can be put into the query string

# Actual Behavior
Other bad things can happen (for example, JS's `alert("hello")`

1. Go to http://codethesaur.us/compare/?concept=data_types&lang1=%3Cscript%3Ealert(%27hello%27)//&lang2=%3C/script%3E

geekygirlsarah:
(Closes Issue #74 )

This resolves injection issues in the query strings discovered by @coreyhaines where JS could be injected into the page. This also resolves a few quirky errors.

image

### DIFF
--- a/web/views.py
+++ b/web/views.py
@@ -1,14 +1,9 @@
-from django.http import (
-    HttpResponse, Http404, HttpResponseRedirect,
-    HttpResponseBadRequest,
-)
-from django.shortcuts import render, get_object_or_404
-from django.urls import reverse
+from django.http import HttpResponse, HttpResponseBadRequest
+from django.shortcuts import render
+from django.utils.html import escape, strip_tags
 
 import json
 import os
-
-# from .models import Question
 
 def index(request):
     with open("web/thesauruses/meta_info.json", 'r') as meta_file:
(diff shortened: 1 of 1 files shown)


web/views.py
```diff
@@ -37,20 +32,20 @@ def compare(request):
...
+
         lang1_directory = meta_data_langs.get(lang1_query_string)
         lang2_directory = meta_data_langs.get(lang2_query_string)
 
-        # if not (lang1_directory and lang2_directory):
-        #     return HttpResponseBadRequest("Lang does not exist")
-
-        concept = request.GET.get('concept', '')
+        if not lang1_directory and lang2_directory:
+            return HttpResponseBadRequest("The " + concept_query_string + " concept of either the " + lang1_query_string + " or " + lang2_query_string + " languages doesn't exist or hasn't been implemented yet.")
```
coreyhaines, line 43: Since we are making some changes around here, what do you think about this returning a 404? It is a missing resource the user is requesting.
`BadRequest : 400` is generally reserved for things like syntax issues in the request.

web/views.py
```diff
@@ -68,10 +63,10 @@ def compare(request):
 
             lang2_friendly_name = lang2_file_json["meta"]["language_name"]
             lang2_categories = lang2_file_json["categories"]
-            lang2_concepts = lang2_file_json[concept]
+            lang2_concepts = lang2_file_json[concept_query_string]
 
     except:
-        return HttpResponse("The " + request.GET.get('concept', '') + " concept of either the " + request.GET.get('lang1', '') + " or " + request.GET.get('lang2', '') + " languages doesn't exist or hasn't been implemented yet.")
+        return HttpResponseBadRequest("The " + concept_query_string + " concept of either the " + lang1_query_string + " or " + lang2_query_string + " languages doesn't exist or hasn't been implemented yet.")
```
coreyhaines, line 69: Same argument for having this be a 404 instead of a 400

web/views.py
geekygirlsarah: That's a good point. Yeah I'll change that.

web/views.py
geekygirlsarah: Also will change!

Issue #125, opened 2020-12-27
Django unit tests run on CI/CD error installing dependencies


geekygirlsarah:
# Expected Behavior
Gitlab Actions runs unit tests and 

# Actual Behavior
Gitlab's PRs just show an error on the "Install Dependencies" tab.

# Steps to Reproduce the Problem

Look at https://github.com/codethesaurus/codethesaur.us/pull/124

geekygirlsarah:
I think this was fixed in #154 (Though I badly worded it and can't 100% tell for sure...)

PR #127, opened 2021-02-10
Resolve install conflicts

cafce25:
With those changes I can get the server running locally on my Linux machine.

The change in pip might be the reason for #125 in which case this should resolve that issue.


geekygirlsarah:
Good catches. Thanks for fixing this! Despite not *actually* using Postgres, it's a required dependency for the Heroku libraries needed.

PR #134, opened 2021-02-25
Docker container for development #132

### LINKED ISSUE #132: Add Docker image
# Description

Some people may not want to install Python and the stuff locally. We should offer a Docker image to run the whole system locally and allow people to work on the project without needing the manual install process.

# Requirements

* Add Dockerfile to build the system. (There may be a Django one out there already, but would need to have some of the extra dependencies added.)
* Update README with directions on how to run it

# Additional Notes

Take a look at https://docs.docker.com/compose/django/

Sarah can look into whether adding it to Docker Hub makes sense or not. If the Dockerfile is super simple, it may not really make sense.

flour-ad:
Pull request contains created Dockerfile and updated README.md file, which now contains instructions on how to run it. 
Repository is provided by mounting the local filesystem, so you can just edit the code in your local IDE and the changes will be reflected in the container. 

To manage the Django app, you have to attach to the container and run the commands within the container. 

During the review, pay attention to the docker image and container names - it is just aesthetic, but maybe you will come up with a better idea.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,3 +3,4 @@ ENV PYTHONBUFFERED=1
 WORKDIR /code
 copy requirements.txt /code/
 RUN pip install -r requirements.txt
+CMD python manage.py runserver 0:8000
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,4 +3,3 @@ ENV PYTHONBUFFERED=1
 WORKDIR /code
 copy requirements.txt /code/
 RUN pip install -r requirements.txt
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3
+ENV PYTHONBUFFERED=1
+WORKDIR /code
+copy requirements.txt /code/
+RUN pip install -r requirements.txt
+
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ with `3`, e.g. `pip3` and `python3`, or follow the process for making Python
 1. Press CTRL+C in the terminal to stop the server
 1. To deactivate the virtual environment, run `deactivate`
 
-### Linux 
+### Linux
 
 1. If python3 and pip3 are not installed there is a guide for [Linux systems](https://www.tecmint.com/install-pip-in-linux/)
 1. Check system default `python --version`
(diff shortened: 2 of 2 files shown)


geekygirlsarah:
This is good, thanks! I appreciate having a Docker container for this!

I was expecting the container to run Django itself, not having to go in and start it yourself. Is there a plus to why you thought running it in Bash might be preferential to just running `docker run` and server is going automatically?

Also I do appreciate the thorough directions! Thanks for that!

flour-ad:
> This is good, thanks! I appreciate having a Docker container for this!
> 
> I was expecting the container to run Django itself, not having to go in and start it yourself. Is there a plus to why you thought running it in Bash might be preferential to just running `docker run` and server is going automatically?
> 
> Also I do appreciate the thorough directions! Thanks for that!

I think I did this because I thought that the user may want to run different commands as well before `runserver` - let me know if you think that implicit launch of the server would be better - I can change that 👍

geekygirlsarah:
> 
> 
> > This is good, thanks! I appreciate having a Docker container for this!
> > I was expecting the container to run Django itself, not having to go in and start it yourself. Is there a plus to why you thought running it in Bash might be preferential to just running `docker run` and server is going automatically?
> > Also I do appreciate the thorough directions! Thanks for that!
> 
> I think I did this because I thought that the user may want to run different commands as well before `runserver` - let me know if you think that implicit launch of the server would be better - I can change that 👍

Yeah, I get that. I think the only main thing one might do is DB migrations, but there's no DB here at the moment. I think in its current state, it should just run the server. (I have plans for adding a DB later but that can be a docker-compose + another command in Docker thing down the line.)

Even if it runs automatically, you can always bypass the command it runs with something like `docker run -it tag_name /bin/bash` to run bash or something else instead.

I guess my aim is for simplicity, especially for people newer to the project.

flour-ad:
OK - now I understand - you are right, as currently there is no need to perform database management, server can be run implicitly. I changed it so that it is activated automatically, and as you said - now when you add "bash" at the end of `docker run ...` command, server launch is skipped and user is able to manually invoke manage.py commands. I committed my changes.

Now I see that some test has failed, but I only edited readme and the Dockerfile. I will check this tomorrow.

geekygirlsarah:
Ignore the failing tests. That's issue #152 and I'm working on fixing it.

geekygirlsarah:
I think this will work! I'll merge it in!

And thank you! I think this will help some people out immensely!

PR #140, opened 2021-02-28
add http to https redirect in non debug envs

### LINKED ISSUE #136: Redirect from HTTP to HTTPS
# Expected Behavior

Going to codethesaur.us shoudl take you to the https site by default.

# Actual Behavior

Going to codethesaur.us may take you to http:// rather than https://. 

1. Type in the URL directly with no http/https on it.

# Specifications

This happens on Firefox and Chrome.

vaibhvshrma:
Fixes #136,
Please verify that it works locally and on prod. I have verified locally. 
Also a gotcha that I found is that browsers can cache redirects made i.e. when I tried going to http with redirects off it didn't redirect (expected), I turned on redirects and went to http and it redirected to https (expected). Now I switched redirects off again and it was still redirecting to https from http. However when I tried this in a new incognito tab for every attempt, all behavior was expected.

### DIFF
--- a/codethesaurus/settings.py
+++ b/codethesaurus/settings.py
@@ -32,6 +32,9 @@ if SYSTEM_ENV == 'PRODUCTION' or SYSTEM_ENV == 'STAGING':
 else:
     DEBUG = True
 
+# redirect all http requests to https in non debugging envs
+SECURE_SSL_REDIRECT = not DEBUG
+
 ALLOWED_HOSTS = []
 
 


geekygirlsarah:
I wasn't expecting this to be so simple! 

I apparently am running Debug mode in production (oops) but this looks correct and I'll go ahead and merge it in.

Thank you!

vaibhvshrma:
Awesome!

PR #157, opened 2021-03-15
Removed first card div in the last card-group row

### LINKED ISSUE #155: Empty box in the bottom of every thesaurus
# Expected Behavior
There should not be an empty box in the bottom of every thesaurus, this box should either contain something or be removed.
# Actual Behavior
There is an empty box in the bottom of every thesaurus in the website, usually to the left of the prompt that calls for people to submit correct information.

1. Go to this link: [Java vs. C++ data types](http://codethesaur.us/compare/?concept=data_types&lang1=java&lang2=cplusplus)
2. Scroll to the bottom of the page

# Specifications
I know this is mostly a nitpick issue and doesn't bother everyone but I think it doesn't look that great. I think we could put a different prompt there or just simply remove it.
* codethesaur.us Version: Latest
* Python Version:3.9

3dvkr:
I had a look at the three boxes, and they get their styles from Bootstrap. That's where its padding/sizing is fixed, and turning off the border for the first box doesn't collapse it away.

I deleted the first card-body from the last row in the html, and things still line up okay. Should I center the links? I'd need to add the `text-center` class to that row--Bootstrap already has this defined. Here are the images comparing the two options.

image
---
image


3dvkr:
Hi @geekygirlsarah, just following up on this. Should I go ahead with centering the text? There's also a similar problem at the end of the reference page ([example](http://codethesaur.us/reference/?concept=data_types&lang=cplusplus)), where there's an empty box in the bottom row. Can I also change that with the same solution/removing that first div entirely?

geekygirlsarah:
I didn't end up looking at this yet, so I'm checking it out now.

I guess I was envisioning something more like this where the columns lined up. So instead of deleting the 3-card row to make a 2-card row (and throwing off alignment), the two cards would be lined up under the columns the languages represent.

Screenshot_2021-03-23 Code Thesaurus - ComparingJava JavaScript

I did this with inspector tools and just turning off the left/top/bottom border of that card, but there might be other ways to do it.

As for centering, I don't know that it makes sense to as nothing else except headers are.

geekygirlsarah:
Hi @3dvkr, I haven't heard back from you on this. If I don't hear back on the comments, I'll just close this so someone else can fix it.

3dvkr:
I ended up removing the borders on the first div in the row, and matched the border properties of the first visible div to the other rows' first div. I think that's the least hack-y way to do it. The class on the first (invisible) div came from Bootstrap, and I added an id for the first visible div's borders. 

The reference.html has the same issue, but the same fix works there: add the border-0 class to the first div, and the start-of-row id to the second div. Should that file be updated as well, or were you thinking of a different layout for that page?

geekygirlsarah:
I like this and think it will work. It fixes it on the compare page, but not on the reference page. Can you apply the same CSS styles on the reference card at the bottom as well? I'll then merge it in.

Thanks for fixing this!

3dvkr:
Hi @geekygirlsarah, the conflict with the branch is with the app.css file. I needed to add a few lines for the #start-of-row id border styling at the very end of the file, but I don't have the permission to resolve the conflict. Other than that, I think this solves issue 155?

geekygirlsarah:
Yes! This works for me! I resolved the conflict and will merge it in.

Thank you for this! Have a great week!

PR #158, opened 2021-03-15
Added the new templates regarding languages

### LINKED ISSUE #139: Add "Report a Language Issue" as issue creation template
# Description

If there's something wrong with a language data set, it's not really a "bug" as much as just "incorrect information." It leaves the template to fill it out kind of weird. This is to revise that template to be a bit more specific to that.

# Requirements

* Add a new template in `.github/ISSUE_TEMPLATE` directory with this
* Have it include prompts like language, language structure, and specific field(s) that are wrong
* Have it optionally say what the correct info is
* Have it also mention if someone also wants to PR it, they need to also leave a comment so it's assigned to them.
* Include people requesting a new language set as well (see below)

# Additional Notes

There should be a place for a person to request a new language structure set too. I'll leave it up to you if this should be in the same template or in a separate template. (Either way, I think both should be included in this PR whether it's 1 or 2 templates.)

See https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository for GitHub's documentation on this.

enisaras:
Closes #139.
I have added two templates, one for language request and one for reporting languages. I wasn't sure where to add the disclaimer concerning the getting assigned to work on it, so I just added it to the bottom without being sure if that is where it should go.

### DIFF
--- a/.github/ISSUE_TEMPLATE/language_issues.md
+++ b/.github/ISSUE_TEMPLATE/language_issues.md
@@ -1,0 +1,24 @@
+---
+name: Language Issue
+about: Use this template to report issues with a language
+---
+
+# Language
+
+<!-- Write out what the language is here -->
+
+# Wrong Field(s)
+1.
+1.
+1.
+<!-- Write out the fields that are wrong here if there are multiple please indicate.-->
+# Corrected Information(Optional)
+1. 
+1. 
+1. 
+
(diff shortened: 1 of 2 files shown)


geekygirlsarah:
I think this works for a starter and can get tweaked if we need to over time.

Thanks! I'll merge it in!

PR #162, opened 2021-03-16
Haskell control structures

cafce25:
Hey,
having control structures probably doesn't hurt Haskell.

I also noticed haskell/data_types.json has execute permissions which it probably doesn't need.


geekygirlsarah:
If they exist in the language, even if they're not conventional or typical, I'd say they should go in here. So I guess confirm that this is all that exists and I'll go ahead and merge it.

As for the permissions... good point. I imagine it's probably something to do with pushing from Windows and/or Mac and/or Linux and some of them mixing it up. Thanks for noticing!

cafce25:
Ok, I've put some research into it and for a part of Haskell (namely Monads) there exist functions that behave like loops. I'll probably have time later today and will add them then.

geekygirlsarah:
This looks good to me. I appreciate the extra research into this!

I'll pull it in.

PR #163, opened 2021-03-16
Add Haskell function syntax

cafce25:
Adding the syntax for Haskell functions.

I realize this looks a little sparse but most concepts just don't make sense in a purely functional language like Haskell.

The only thing I am a little unsure about is the call_subroutine one is that just meant to be function invocation? In that case I would add it.


geekygirlsarah:
I don't think Haskell has subroutines so this is fine.

Thanks! I'll go ahead and merge it in.

PR #166, opened 2021-03-16
Respect newlines and indentation of concept files

### LINKED ISSUE #156: Python code is indentation sensitive and should be shown with indentation
I suspect this is a general issue with the front-end and applies to other indentation sensitive languages too. 

# Expected Behavior
The control structures for python are displayed with correct indentation like so:
```python
if condition:
    statements
```
# Actual Behavior
The indentation gets stripped out by the front-end resulting in
```
if condition:
statements
```

1. Visit [The control structures page for python](https://codethesaur.us/reference/?concept=control_structures&lang=python)

# Specifications
* codethesaur.us Version: git sha 'dae0b20'
* Python Version: 3.9.2

cafce25:
I think treating the code snippets as pre-formatted text with `<pre>` is an appropriate way of resolve #156 

The css changes are necessary to keep the look and feel.

A detail that is changed by this is that now long lines no longer automatically wrap, but I don't know if we want that to happen or if sideways scrolling is fine in that case.

### DIFF
--- a/web/static/css/app.css
+++ b/web/static/css/app.css
@@ -108,6 +108,10 @@ footer p {
   background-color: #fff;
 }
 
+.card-body pre {
+  color: #e83e8c;
+}
+
 a.toTop {
   color: #304ffe;
   text-decoration: none;
--- a/web/templates/comparecard.html
+++ b/web/templates/comparecard.html
@@ -2,7 +2,7 @@
     <div class="card-body">
         <div>
             {% if code %}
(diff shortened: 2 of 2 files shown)


cafce25:
One could alternatively solve this with a css rule for the code blocks: `white-space: pre;` i didn't go that route because I did not know better when I created this but it's an easy edit if that solution is preferred.

geekygirlsarah:
I think this looks good and seems to work in the review app. I'll pull it in!

Thank you!

PR #178, opened 2021-04-13
Add a step to copy all files to the docker environment

sophiedeziel:
The dockerfile was missing a step to be able to run the project.


geekygirlsarah:
This is lovely! Thank you for fixing this!

Issue #145, opened 2021-02-28
Update Heroku instance of CT


geekygirlsarah:
Due to it being on Heroku, only Sarah can do all of this.

* Update Python to latest version
* Update Django to latest version
* Update django-heroku module to django-on-heroku
* Turn SSL on and configure static files to serve correctly (disable collectstatic and enable whitenoise?)
* Fix SSL and Debug settings so they're correct for each given environment. (maybe default SSL false and Debug true unless env variable is set?)

https://pypi.org/project/django-on-heroku/

https://stackoverflow.com/questions/17482659/how-to-serve-django-static-files-on-heroku-with-gunicorn

geekygirlsarah:
Tried to do this, `DEBUG = False` then means static files aren't served up. According to https://devcenter.heroku.com/articles/django-assets , all the things it suggest have already been implemented.

If anyone knows the solution, I'll gladly take it!

geekygirlsarah:
Fixed with a9f2f6a, f2f0fd6, f26c71b, 3b866f5, 7086a66, 2362a8d, 78e9616.

PR #179, opened 2021-04-13
Upgrade Django from 3.1.6 to 3.2

sophiedeziel:
Part of #145
Overrides #176

Upgrading Django from 3.1.6 to 3.2


geekygirlsarah:
Woo hoo! upgrades! Thank you for this! This is very appreciated!

PR #181, opened 2021-04-14
Add Go control structures

### LINKED ISSUE #107: Add Go control structures
# Description

The control structures (loops, ifs, etc.) in Go are presently not on Code Thesaurus. I would love to see this added.

# Requirements

* Copy `web/thesauruses/_meta/control_structures.json` to a `web/thesauruses/go/` directory
* Edit the control_structures.json file to accurately reflect Go and the latest major version of it.

# Additional Notes

Please leave comments on this issue to ask questions or to claim the issue as yours. I'll keep it yours for a week then remove you in case someone else wants to work on it. (You can always reclaim it.)

Check out [CONTRIBUTING.md](CONTRIBUTING.md) for more information on how to contribute. 

Check out [https://docs.codethesaur.us/thesaurus/](https://docs.codethesaur.us/thesaurus/) for information on more specifics of implementing this.

kellydanma:
Adds Go control structures 🎉

### DIFF
--- a/web/thesauruses/_meta/control_structures.json
+++ b/web/thesauruses/_meta/control_structures.json
@@ -10,8 +10,7 @@
       "if_else_conditional",
       "if_elseif_conditional",
       "if_elseif_else_conditional",
-      "ternary_conditional",
-      "defer"
+      "ternary_conditional"
     ],
     "Loops": [
       "while_loop",
--- a/web/thesauruses/_meta/control_structures.json
+++ b/web/thesauruses/_meta/control_structures.json
@@ -10,7 +10,8 @@
       "if_else_conditional",
       "if_elseif_conditional",
       "if_elseif_else_conditional",
-      "ternary_conditional"
+      "ternary_conditional",
+      "defer"
     ],
     "Loops": [
       "while_loop",
@@ -47,6 +48,10 @@
     },
     "ternary_conditional": {
       "name": "Ternary conditional",
+      "code": ""
+    },
+    "defer": {
+      "name": "Defer statement",
       "code": ""
     },
     "while_loop": {
--- a/web/thesauruses/_meta/control_structures.json
+++ b/web/thesauruses/_meta/control_structures.json
@@ -49,10 +49,6 @@
       "name": "Ternary conditional",
       "code": ""
     },
-    "defer": {
-      "name": "Defer statement",
-      "code": ""
-    },
     "while_loop": {
       "name": "While loop",
       "code": ""
--- a/web/thesauruses/go/control_structures.json
+++ b/web/thesauruses/go/control_structures.json
@@ -9,7 +9,8 @@
       "if_conditional",
       "if_else_conditional",
       "if_elseif_conditional",
-      "if_elseif_else_conditional"
+      "if_elseif_else_conditional",
+      "defer"
     ],
     "Loops": [
       "for_loop",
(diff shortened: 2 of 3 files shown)


kellydanma:
@geekygirlsarah I noticed we don't have keys for `case` or `switch` statements. I think these belong in control structures. I don't have to add one now for Go, but I can do so in an upcoming PR. Let me know what you think!

web/thesauruses/_meta/control_structures.json
```diff
@@ -10,7 +10,8 @@
       "if_else_conditional",
       "if_elseif_conditional",
       "if_elseif_else_conditional",
-      "ternary_conditional"
+      "ternary_conditional",
+      "defer"
```
kellydanma: `defer` isn't idiomatic in most languages, but is a crucial part of Go. That's why I added it; if we don't want to add it for now, let me know.

web/thesauruses/_meta/control_structures.json
```diff
@@ -94,4 +99,4 @@
       "code": ""
     }
   }
-}
\ No newline at end of file
+}
```
kellydanma, line 97: Might as well fix the missing newline while I'm here 😄

geekygirlsarah:
> @geekygirlsarah I noticed we don't have keys for `case` or `switch` statements. I think these belong in control structures. I don't have to add one now for Go, but I can do so in an upcoming PR. Let me know what you think!

That's issue #186!

geekygirlsarah:
I'll review this soon! Thanks for doing it!

web/thesauruses/_meta/control_structures.json
cafce25: I'm a bit unsure if defer is in the right place here. How i understand them control structures are used to determine which pieces of code are executed.
While defer changes when code is run, apart from terminating the program early it doesn't change if a piece of code is run.
I think something like defer would be better placed somewhere else, for example in functions like [An Introduction to Programming in Go](http://www.golang-book.com/books/intro) and others put it.

kellydanma: I'm okay with that too. The official tour of Go _does_ put it under control structures FYI: https://tour.golang.org/flowcontrol/12

I'm not strongly opinionated about this.

kellydanma: After further thought, I would have to disagree that `defer` belongs in functions. It is, after all, still a control flow `keyword`. And you may use multiple of them. I think this article explains the role of `defer` pretty well: https://www.digitalocean.com/community/tutorials/understanding-defer-in-go

I realize it's not "traditional" in the sense of whether or not something gets executed.

geekygirlsarah: First, I greatly appreciate these discussions! I'm familiar with concepts of Go but haven't really used it, so this is insightful.

Second, I looked it up. It reminds me of an `await` statement in JS. So the idea at least belongs in more than one language and isn't Go-exclusive. (Though upon reading more, stacking defers goes in reverse order, which is interesting.)

I'm rolling this around in my head and think I could see it more in conditionals. While it really only works for calling functions, it's not a part of defining those functions. (Though also in JS, when you call `await` before a function, the function has to also have `async` in front of it, which _would_ be a part of defining the function.) So really it could perhaps belong in both. But for now I think it could belong here.

Since there's no `defer` ID yet, and the language matching code isn't smart enough to combine these together yet when various IDs are missing or there's extras, I wouldn't add it yet. Optionally you can add it and test it to see if it will work.

cafce25: After reading your helpful documentation and how Swift uses `defer` I'm convinced you're right and `defer` does indeed belong in `control_structures.json`.

The category "Conditional Statements (ifs)" still feels off though.

I've looked into it a little and think we might need a category for `try`...`finally` & co. maybe `defer` fit's in there since it's used in a very similar fashion? I've created an issue #194 for the new category. Maybe we can just leave `defer` here until that issue gets resolved?

geekygirlsarah: I had pondered if try/catch/finally type things belonged in a Error Handling structure (like in the drop down on the home page). This has me pondering if it belongs here. (I want to say "why not both?" but could mean one could get updated without the other.)

I could see, at least for now, an Error Handling category added and adding several things under there. How's that sound?

kellydanma: I am aligned on that, removed 👍

geekygirlsarah:
This looks great to me! I'm happy to merge it in!

Thank you, I appreciate it!

PR #183, opened 2021-04-18
Docker Setup for Code Thesaurus

ChaelCodes:
# Description
The current setup has a `docker run` command with two steps for building the image and then running the image with volumes and the like. We'd like to simplify this into one `docker-compose up command`

# Changes
- Add a `docker-comopse.yml` to simplify and standardize the build process
- Name the Container `cthesaurus-img` so it can be referenced from `docker-compose.yml`


geekygirlsarah:
This looks great to me! I appreciate the addition!

I'll merge it in!

PR #191, opened 2021-04-18
Correct Spelling of conditional in Ruby

ChaelCodes:
# description
Correct spelling of Conditional in Ruby Control structures.


geekygirlsarah:
Good catch @ChaelCodes and @SteveVaneeckhout! I'll merge it in!

PR #195, opened 2021-04-28
Add syntax highlighting using pygments

### LINKED ISSUE #170: Add a highlighting library for formatting/coloring
# Description

This would add a code highlighting library so that code objects would be formatted and colored according to the style of the language displayed.

# Requirements

1. Research Prism.js, Highlight.js, or Pygments to see what might be the best scenario to support the most languages and easily integrate into CT.
1. Add the library to the main header/footer templates, then in appropriate compare/reference templates.
1. Ensure the column of that language is formatted correctly for that particular language. (This may mean passing over another variable to the template from the controller/view if need be.)

# Additional Notes

If you think there's a better syntax highlighter than these, feel free to drop a comment and I'd be happy to look it over.

cafce25:
This adds syntax highlighting using pygments with the `colorful` style.

Close #170

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ isort==5.5.4
 lazy-object-proxy==1.4.3
 mccabe==0.6.1
 psycopg2-binary==2.8.6
+Pygments==2.8.1
 pylint==2.6.0
 pytz==2021.1
 six==1.15.0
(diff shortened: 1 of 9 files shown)


geekygirlsarah:
This looks great to me! I'll merge it in.

Thank you again for the great work!

PR #196, opened 2021-04-28
Add Ada thesauruses

cafce25:


geekygirlsarah:
Adding Ada! So cool! A new language in CT now!

Thanks yet again for more great work!

PR #210, opened 2021-05-04
Added 'functions' entries for Scala for issue #103.

### LINKED ISSUE #103: Add Scala functions
# Description

The functions in Scala are presently not on Code Thesaurus. I would love to see this added.

# Requirements

* Copy `web/thesauruses/_meta/functions.json` to a `web/thesauruses/scala/` directory
* Edit the functions.json file to accurately reflect Scala and the latest major version of it.

# Additional Notes

Please leave comments on this issue to ask questions or to claim the issue as yours. I'll keep it yours for a week then remove you in case someone else wants to work on it. (You can always reclaim it.)

Check out [CONTRIBUTING.md](CONTRIBUTING.md) for more information on how to contribute. 

Check out [https://docs.codethesaur.us/thesaurus/](https://docs.codethesaur.us/thesaurus/) for information on more specifics of implementing this.

hypatian:
Possibly more detail than is needed and some of this should be
split into different concepts at some point.

- [x] Data update

## 📖 Description

Added entries for all of the applicable functions concepts for Scala 2.13 (issue #103). Scala has a couple of things that don't really fit into the existing concepts, and I'm uncertain about whether I should leave those details out or put them in.

### DIFF
--- a/web/thesauruses/scala/functions.json
+++ b/web/thesauruses/scala/functions.json
@@ -1,0 +1,71 @@
+{
+  "meta": {
+    "language": "scala",
+    "language_version": "2.13.x",
+    "language_name": "Scala"
+  },
+  "categories": {
+    "Void Functions": [
+      "void_function_no_parameters",
+      "void_function_with_parameters",
+      "void_function_variable_parameters"
+    ],
+    "Return Value Functions": [
+      "return_value_function_no_parameters",
+      "return_value_function_with_parameters",
+      "return_value_function_variable_parameters"
+    ],
+    "Lambda/Anonymous Functions": [
+      "anonymous_function_no_parameters",
+      "anonymous_function_with_parameters",
+      "anonymous_function_variable_parameters"
+    ],
+    "Subroutines": [
+      "call_subroutine",
+      "return_from_subroutine"
+    ]
+  },
+  "functions": {
+    "void_function_no_parameters": {
+      "code": "// For simple expressions, simply give the expression:\ndef methodName(): Unit = expression\n\n// For more complex expressions, or to make local definitions,\n// use a block instead:\ndef methodName(): Unit = {\n  expressions and definitions\n}",
+      "comment": "Blocks (which may contain multiple expressions and definitions and evaluate to the value of the last expression) are a kind of expression."
+    },
+    "void_function_with_parameters": {
+      "code": "// You may have a single parameter list with multiple parameters:\ndef methodName(param1: Type1, param2: Type2, ...): Unit =\n  expression\n\n// You may also have zero or more parameter lists each with\n// zero or more parameters:\ndef methodName(param1a: Type1a, ...)\n              (param1b: Type1b, ...)...: Unit = expression"
+    },
+    "void_function_variable_parameters": {
+      "code": "// In each parameter list, a single extra variable parameter may\n// follow defined parameters:\ndef methodName(param1: Type1,..., extraParams: ExtraType*): Unit =\n  expression",
+      "comment": "`extraParams` will have type `Seq[ExtraType]`. Note that multiple parameter lists may each have their own extra parameters."
+    },
+    "return_value_function_no_parameters": {
+      "code": "// For simple expressions, simply give the expression:\ndef methodName(): ResultType = expression\n\n// For more complex expressions, or to make local definitions,\n// use a block instead:\ndef methodName(): ResultType = {\n  expressions and definitions\n}\n\n// You may also have zero parameter lists:\ndef methodName: ResultType = expression",
+      "comment": "A method with zero parameter lists is referenced like accessing a field (`object.methodName`), but executes code every time it is evaluated. Such methods are typically used for methods that are essentially computed fields with no side-effects."
+    },
+    "return_value_function_with_parameters": {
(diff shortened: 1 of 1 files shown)


hypatian:
About maybe getting too verbose: I've included a few different examples of several cases.

In Scala, everything is an expression, but a block in curly braces is a common sort of expression, so I have included some examples to make both of those things obvious to new Scala users. This might be better served with some sort of "how do you put multiple expressions together into a single expression" or "how do you group multiple executable lines of code" or similar concept in the future.

Most scala functions are _written_ as methods, which is what I've shown, with the syntax for function values under "anonymous functions". This seems fine and matches the way some other languages have done this.

Scala methods may have a variable number of _parameter lists_, in addition to having a variable number of parameters in each list. I've included these as examples in definitions of multiple-argument functions. This also includes the difference between `def foo() ...` for a method that takes a single empty parameter list, and `def bar ...` for a method that takes _zero_ parameter lists. I've also included both of these in examples.


So, questions: Obviously this is more comprehensive than it perhaps needs to be. But, without some of the explanation readers may be confused when they see these common features used in actual code. And in fact, the multiple parameter lists feature shows up in the control structures (the `fold` methods take argument lists.)

What's the right balance of thorough and simple/clean? Especially for niche features that nearly no other language has. Some of these things should perhaps be split later into multiple concepts. But until they are, if they ever are, does it make sense to include them in these concepts?

geekygirlsarah:
This is pretty close! You mostly figured it out.

`"name"` is a field for the meta files and no longer needed in language files. You can remove all of those (though it won't break anything to leave those in.)

For grouping/blocking expressions, I can see that going in another category. No need to explain that here.

If something's not _technically_ write but you can kind of hack something together to make it act like it, I'm fine leaving it out, but if left in, I think leave a `"comment":` for it. You seemed to do that.

For multiple examples, I thoughta bout making `"code"` an array rather than a string, but haven't gotten that far yet. I don't mind multiple examples in there, but maybe either mention in `"comment"` it's two things or maybe add a comment showing like `-- Example 1:` and `-- Example 2:` kind of thing.

As for balance of simple/clean: It's hard. I don't necessarily have that yet. I see this as a "reference" tool, not an "educational" tool. (As in you use it for quickly looking up things, not being a textbook that explains every feature everywhere. If they need a textbook, there's textbooks.) At some point I'll get a style guide up but I'm not quite there yet.

As for niche features... I'm perfectly fine with "you can do this in X language, you can't in Y language" kind of things. I built that into CT. All IDs need to be defined in the files, but if they don't exist in the language, just do `"not-implemented": true` instead of code. You can also include a comment (either why it's not there or an alternative). These will show up differently than if a code sample was included.

Did I get all the questions? 😁

hypatian:
Ahh! I see the not-implemented snuck by me while I was working on this. I've cleaned up for that.

I tweaked the code:

I've tried to make it so that each outside not-yet-documented concept is only included once per section of the output (and I recognize that this isn't the best because the way the output is organized may change in the future.) I'm aiming for "enough that someone can get by." Seeing just `def methodName(): Unit = expression` doesn't give any idea that braces are used for grouping, so I wanted to do more than that.

I figure some of the extra things can be split into additional concepts in the future, and at that time some of the multiple examples will be removed from sections that include them.

I've also included some markdown in the comments. It's not being processed, but I noted that it's present in some of the comments from other languages (like Ruby, for example). This feels like something that ought to be supported in the future, so I've followed that example.

I'm happy to make any changes to further trim this down or change the way I've done things, however. Just let me know what you want!

geekygirlsarah:
I added #213 for the Markdown.

And scroll to the bottom of the To-Do column on https://github.com/orgs/codethesaurus/projects/4 if you want to see other structures I intend to add in the future.

I'll review this probably in the next day or so, along with your other one. But I appreciate the contribution so far! And fun to see another person I know work on this project!

hypatian:
> I'll review this probably in the next day or so, along with your other one. But I appreciate the contribution so far! And fun to see another person I know work on this project!

Thanks! Happy to help out. ^_^

web/thesauruses/scala/functions.json
```diff
@@ -0,0 +1,71 @@
...
+      "anonymous_function_variable_parameters"
+    ],
+    "Subroutines": [
+      "call_subroutine",
+      "return_from_subroutine"
+    ]
+  },
+  "functions": {
+    "void_function_no_parameters": {
+      "code": "// For simple expressions, simply give the expression:\ndef methodName(): Unit = expression\n\n// For more complex expressions, or to make local definitions,\n// use a block instead:\ndef methodName(): Unit = {\n  expressions and definitions\n}",
```
geekygirlsarah, line 30: You don't have to describe what's happening in comments. The idea is to try to keep the code clean, and move any clarifications into `comment`.

(I can still merge it, just it makes it more bloated to compare code samples had might end up just more confusing)

geekygirlsarah:
The failed deployment is due to Django `settings.py` changes that break any build that hasn't had those changed pulled in yet. This will still work fine when merged.

I'll go ahead and merge it in, though you may want to clear up the comments out of the code and/or add more to comments. 

Thank you for adding this, I appreciate it!

PR #240, opened 2021-08-13
Add classes for Nim

juancarlospaco:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## 📖 Description

- Add `classes.json` for http://nim-lang.org

## ✅ QA Instructions, Screenshots

Nim explained a little bit:

- Nim does not use Class, it uses Types instead, Types are objects.
- Nim uses "Static Dispatch".
- All Types are Static. All Functions are Static. All Attributes are Static.
- Object methods are just free floating functions.
- At compile-time functions are "glued" to objects.
- All Symbols are private by default, if a symbol has an asterisk on the declaration is Public, then `var foo = 42` is Private and `var foo* = 42` is Public.
- All Symbols are immutable by default, if a symbol has a `var` declaration is Mutable, then `let foo = 42` is immutable and `var foo = 42` is mutable.
- Type instantiation is just `Name()`, but if you need a custom Constructor, any function that returns the Type is a Constructor.

🙂

### DIFF
--- a/web/thesauruses/nim/classes.json
+++ b/web/thesauruses/nim/classes.json
@@ -1,0 +1,127 @@
+{
+  "meta": {
+    "language": "nim",
+    "language_version": "1.4",
+    "language_name": "Nim"
+  },
+  "categories": {
+    "Defining Classes": [
+      "normal_class",
+      "abstract_class",
+      "interface",
+      "read_only_class",
+      "static_class",
+      "inner_class"
+    ],
+    "Adding Private and Public Members": [
+      "private_variables",
+      "protected_variables",
+      "public_variables",
+      "static_variables",
+      "private_functions",
+      "protected_functions",
+      "public_functions",
+      "static_functions"
+    ],
+    "Extending and Implementing Classes": [
+      "extends_class",
+      "extending_interface",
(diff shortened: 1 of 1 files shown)


web/thesauruses/nim/classes.json
```diff
@@ -0,0 +1,127 @@
...
+      "name": "Interface",
+      "code": "type Name = concept"
+    },
+    "read_only_class": {
+      "name": "Read-only class",
+      "code": "type Name = object"
+    },
+    "static_class": {
+      "name": "Static class",
+      "code": "Types are static"
```
geekygirlsarah: Should this be `"comment": "Types are static"` instead?

juancarlospaco: Fixed.

geekygirlsarah:
You can remove all the `"name": "..."` parts from this file. They're only needed in the meta language file.

Fix that and the comment and I'll merge it in!

geekygirlsarah:
Sweet! I'll merge it in. Thank you again for your contributions!

PR #244, opened 2021-08-15
Issue 242 fix 404 error pages

### LINKED ISSUE #242: 404 error pages sometimes have incorrect descriptive text
# Expected Behavior

For issues of a concept and language not being implemented in a .json file yet, the page should say that.

# Actual Behavior

It says the language isn't valid.

1. On the home page, choose something that doesn't exist, like Classes, C++, and Scala. (At least, as of Aug 14)
1. The resulting page says "The first language (cpp) isn't valid"

# Additional Info

This will affect `/compare` and `/reference` pages, both of which are in the `web/views.py` file.

felis-blue:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## 📖 Description

Fixes #242

Adds a new method `lang_exists` to the Language class, that checks if the path for the language exists in the thesauruses folder.
Uses this method to determine if a language exists, if so a different error message is displayed.

## ✅ QA Instructions, Screenshots

Modify the lang parameters in the urls for /compare and /reference:
* Same behavior as before when the language does not exist.
* New error message if the language exists but doesn't have an entry for the concept.

### DIFF
--- a/web/Language.py
+++ b/web/Language.py
@@ -24,6 +24,13 @@ class Language:
         # Empty string is falsy, but text is truthy, but would return return text
         return bool(self.key)
 
+    def lang_exists(self):
+        """
+        Returns a Boolean if the language (self.key) exists in the thesauruses or not
+        :rtype: bool
+        """
+        return os.path.exists(os.path.join("web", "thesauruses", self.key))
+
     def load_structure(self, structure_key):
         """
         Loads the structure file into the Language object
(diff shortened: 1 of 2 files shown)


felis-blue:
This is what I came up with to solve the issue, I hope it makes sense! Let me know if you want anything changed!

geekygirlsarah:
That's a great fix! I'll merge it in. Thank you!

PR #353, opened 2021-10-10
Updated c# control structures.

### LINKED ISSUE #342: C# Switch statement
# Language
C#

# Wrong Field(s)
1. Missing switch statement

# Corrected Information(Optional)
This would be a standard switch:
switch(var) {
    case condition:
        statements;
        break;
    default:
        statements;
        break;
}

However C# has both the switch statement and switch expression which does a similar task. The second use case is more for pattern matching.

blockingHD:
 
Resolves #342 

## What changed and why?

Added the additional language features from the main meta file for control structures.

## Checklist

Added in the use of System.Linq to achieve some of the functional elements such as map and reduce. Given it is part of the core assembly for .Net I think it serves its purpose in the context but it is a framework feature and not strictly a C# feature therefore wouldn't be available in Mono etc etc.

Only included the switch statement in this as other implementations in the language are focused around pattern matching which should probably be a concept in its own right.

### DIFF
--- a/web/thesauruses/csharp/control_structures.json
+++ b/web/thesauruses/csharp/control_structures.json
@@ -31,35 +31,27 @@
   },
   "control_structures": {
     "if_conditional": {
-      "name": "If conditional",
       "code": "if (condition)\n{\n     statements;\n}"
     },
     "if_else_conditional": {
-      "name": "If/Else conditional",
       "code": "if (condition)\n{\n     statements;\n} \nelse\n{\n     statements;\n}"
     },
     "if_elseif_conditional": {
-      "name": "If/ElseIf conditional",
       "code": "if (condition)\n{\n    statements;\n} \nelse if (condition)\n{\n    statements;\n}"
     },
     "if_elseif_else_conditional": {
-      "name": "If/ElseIf/Else conditional",
       "code": "if (condition)\n{\n    statements;\n} \nelse if (condition)\n{\n    statements;\n}\nelse\n{\n    statements;\n}"
     },
     "switch_statement": {
-      "name": "Switch statement",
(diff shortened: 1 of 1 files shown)


web/thesauruses/csharp/control_structures.json
```diff
@@ -54,29 +62,42 @@
...
+    "until_loop": {
+      "not-implemented": true
+    },
+    "do_until_loop": {
+      "not-implemented": true
+    },
     "for_loop": {
       "name": "For loop",
-      "code": "for(variables; conditions; expression after iteration)\n{\n    statements\n}"
+      "code": "for(variables; conditions; expression after iteration)\n{\n    statements;\n}"
```
geekygirlsarah: Can you change "expression after iteration" to something like "expressionAfterIteration"? I'm trying to make sure all the code blocks are something equivalent to actually compileable code.

geekygirlsarah (changes requested):
Change that one thing and I'll merge this in!

web/thesauruses/csharp/control_structures.json
```diff
@@ -29,9 +31,11 @@
   },
   "control_structures": {
     "if_conditional": {
+      "name": "If conditional",
       "code": "if (condition)\n{\n     statements;\n}"
     },
     "if_else_conditional": {
+      "name": "If/Else conditional",
```
geekygirlsarah: You should be able to delete the `name` lines since they exist in the meta file. (I think I didn't catch this before, sorry.)

geekygirlsarah (changes requested):

geekygirlsarah:
Thanks for the changes. It looks right, so I'll merge it in.

Don't forget to sign up for Hacktoberfest if you haven't yet!

PR #354, opened 2021-10-10
Issue 309 add rust logical and math operators

### LINKED ISSUE #309: Add Rust Logic/Math Operators
# Description

Logic/Math Operators for Rust aren't presently on Code Thesaurus. This issue is to add that concept to the system.

# Requirements

* Copy `web/thesauruses/_meta/operators.json` to the right Rust directory in `web/thesauruses/`
* Edit the operators.json file to accurately reflect the latest major version of Rust
* Add in `"not-implemented": "true"` if the concept doesn't exist in the language. You can then leave `"code":` blank.

# Claiming This Issue

# Need Help?

# Thanks!

bug-author:

Resolves #309 

## What changed and why?

Added Logical and Math operators for Rust 

## Checklist

I renamed my branch to fix a typo before pushing the final commit, I guess that's the reason GitHub is showing commits as if they are from multiple users in this PR.

### DIFF
--- a/web/thesauruses/rust/operators.json
+++ b/web/thesauruses/rust/operators.json
@@ -110,14 +110,14 @@
       "not-implemented": true
     },
     "exponential": {
-      "not-implemented": true,
+      "code": ["pow(number, exponent)", "checked_pow(number, exponent)"],
       "comment": "Rust provides methods pow and checked_pow for exponentiation."
     },
     "factorial": {
       "not-implemented": true
     },
     "absolute_value": {
-      "not-implemented": true,
+      "code": "number.abs()",
       "comment": "Rust provides abs method for finding absolute value."
     },
     "percentage": {
--- a/web/thesauruses/rust/operators.json
+++ b/web/thesauruses/rust/operators.json
@@ -1,0 +1,241 @@
+{
+  "meta": {
+    "language": "language_id",
+    "language_version": "version.number",
+    "language_name": "Human-Friendly Language Name"
+  },
+  "categories": {
+    "Arithmetic Operators": [
+      "addition",
+      "addition_assignment",
+      "subtraction",
+      "subtraction_assignment",
+      "multiplication",
+      "multiplication_assignment",
+      "division",
+      "division_assignment",
+      "integer_division",
+      "integer_division_assignment",
+      "modulus",
+      "modulus_assignment",
+      "unary_plus",
+      "unary_minus",
+      "increment",
+      "decrement",
+      "exponential",
+      "factorial",
+      "absolute_value",
+      "percentage"
+    ],
+    "Equality/Comparison Operators": [
+      "equal_to",
+      "not_equal_to",
+      "less_than",
+      "less_than_or_equal_to",
+      "greater_than",
+      "greater_than_or_equal_to",
+      "null_coalescing",
+      "is",
+      "is_not"
+    ],
+    "Logical/Boolean/Bitwise Operators": [
+      "and",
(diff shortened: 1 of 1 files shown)


web/thesauruses/rust/operators.json
```diff
@@ -0,0 +1,203 @@
...
+    },
+    "increment": {
+      "not-implemented": true
+    },
+    "decrement": {
+      "not-implemented": true
+    },
+    "exponential": {
+      "not-implemented": true,
+      "comment": "Rust provides methods pow and checked_pow for exponentiation."
```
geekygirlsarah, line 114: could you give an example of both? Sounds like it's implemented as a function rather than an operator (but it'll work in this case)

web/thesauruses/rust/operators.json
```diff
@@ -0,0 +1,203 @@
...
+    },
+    "factorial": {
+      "not-implemented": true
+    },
+    "absolute_value": {
+      "not-implemented": true,
+      "comment": "Rust provides abs method for finding absolute value."
```
geekygirlsarah, line 121: Same for `abs` here.

geekygirlsarah (changes requested):
I think the two small changes and this is good to merge!

web/thesauruses/rust/operators.json
bug-author: Sure. Just to make it clear, I should remove the `not-implemented` key and provide both examples as a part of `code`?

geekygirlsarah: Yeah. I don't know the difference between the two but if they both serve some form of doing exponentials, then I'd like to have them here! Despite this being operators, I don't know that I imagine any other place to put functions like that in CT.

bug-author:
@geekygirlsarah I've made the changes as per the requirements.

web/thesauruses/rust/operators.json
geekygirlsarah: To the best of my knowledge, `code` only is a string, not an array/list. Did you test this to confirm it works still?

bug-author: I have now added both methods as a single string separated by comma, and tested the code.

I was confused as there were two methods that's why I asked if I should provide both of them as a part of `code`. There are no multi-line strings allowed in JSON, so arrays seemed semantically they correct way for two different entities belonging to one key. I should have tested the code after making changes, I'm so sorry.

geekygirlsarah: Typically people have added `\n` to separate out new lines in JSON strings, which is probably what I'd prefer here.

And yeah, an array/list does tend to make sense, but that's sort of where looking over docs or seeing how others do it would come in handy. It might be a thing we can add, but it's not how the system works presently (though, like I said, I never tested it so I'm not sure what it would do).

bug-author: I've made the changes and tested with `\n`. Once again, I apologize for being hasty and not looking into docs - I'm going through them now!

I'm a newbie dev so I might be wrong, but I think the array/ list approach would result in better readability - presently it does not work I tested it.

geekygirlsarah:
> 
> 
> @geekygirlsarah I've made the changes as per the requirements.

For future reference, you can always click "Resolve conversation" to notify the maintainer you've made the changes on a PR.

Also I think it looks better. I'll merge it in.

Thanks! This counts towards Hacktoberfest, so be sure to sign up if you want this to count towards the shirt/swag!

PR #368, opened 2021-10-14
Added string.json to c# and started compleating

### LINKED ISSUE #365: Add C# Strings
# Description
Strings for C# aren't presently on Code Thesaurus. This issue is to add that concept to the system.

# Requirements
Copy web/thesauruses/_meta/strings.json to the right C# directory in web/thesauruses/
Edit the strings.json file to accurately reflect the latest major version of C#
Delete the "name": fields before your PR
Add in "comment": if you wish to explain something in the code
Add in "not-implemented": "true" if the concept doesn't exist in the language. You can then leave "code": blank.

blockingHD:

Resolves #365

## What changed and why?
Added strings.json to C#.

## Checklist

Using `str` as `string` is a keyword
Preferring `string` over `String`

### DIFF
--- a/web/thesauruses/csharp/strings.json
+++ b/web/thesauruses/csharp/strings.json
@@ -1,0 +1,342 @@
+{
+  "meta": {
+    "language": "language_id",
+    "language_version": "version.number",
+    "language_name": "C#"
+  },
+  "categories": {
+    "Creating/Destroying Strings": [
+      "is_primitive_or_not",
+      "import",
+      "default_string_byte_encoding",
+      "create_new_string",
+      "create_multiline_string",
+      "assign_new_string",
+      "destroy_string"
+    ],
+    "String Basics": [
+      "length_of_string",
+      "max_length_of_string",
+      "clear_string",
+      "is_empty",
+      "concatenate_two_strings",
+      "concatenate_many_strings",
+      "is_all_alphabetical",
+      "is_all_numerical",
+      "is_all_alphanumeric",
+      "is_decimal",
+      "is_all_whitespaces",
+      "is_all_uppercase",
+      "is_all_lowercase",
+      "is_in_titlecase"
+    ],
+    "Find and Search": [
+      "does_substring_exist",
+      "find_start_index_of_substring",
+      "find_start_index_of_additional_substring",
+      "find_start_index_of_substring_from_end",
+      "count_occurrences_of_substring",
+      "get_leftmost_characters",
+      "get_rightmost_characters",
+      "get_substring_from_start_and_end_index",
+      "get_substring_from_start_index_and_length"
+    ],
+    "Manipulating Strings": [
+      "convert_to_uppercase",
+      "convert_to_lowercase",
+      "convert_to_title_case",
+      "capitalize_string",
+      "remove_whitespace",
+      "replace_substring",
+      "replace_all_substring",
+      "split_at_index",
+      "split_at_newlines",
+      "split_at_substring",
+      "merge_lists_into_string",
+      "encode_html_entities",
+      "decode_html_entities",
+      "encode_url_percent",
+      "decode_url_percent",
+      "encode_to_base64",
+      "decode_from_base64"
+    ],
+    "String Formatting": [
+      "format_string_function",
+      "parameter_format_in_order",
+      "parameter_format_numerical",
+      "parameter_format_by_name",
+      "format_as_integer",
+      "format_as_decimal",
+      "format_as_fixed_decimal",
+      "format_as_currency",
+      "format_as_percentage",
+      "format_number_with_separators",
+      "format_number_with_positive_negative_sign",
+      "format_number_in_scientific_notation_little_e",
+      "format_number_in_scientific_notation_big_e",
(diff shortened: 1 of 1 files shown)


web/thesauruses/csharp/strings.json
```diff
@@ -0,0 +1,342 @@
+{
+  "meta": {
+    "language": "language_id",
+    "language_version": "version.number",
+    "language_name": "C#"
```
geekygirlsarah: The language ID is "csharp" (like the folder it's in). the version I think is 9.0? Or use whatever the last major.minor version is.

web/thesauruses/csharp/strings.json
```diff
@@ -0,0 +1,342 @@
...
+      "code": "using  System;",
+      "comment": "string will be avalible without this use but String will not be."
+    },
+    "default_string_byte_encoding": {
+      "name": "Default byte encoding (ex: ASCII, UTF-8, UTF-16, etc.)",
+      "code": "UTF-16"
+    },
+    "create_new_string": {
+      "name": "Create new string",
+      "code": "var newString = \".... . .-.. .-.. ---\""
```
geekygirlsarah: Obscure, but I appreciate this as a licensed ham radio operator, lol

web/thesauruses/csharp/strings.json
```diff
@@ -0,0 +1,342 @@
...
+      "format_number_in_scientific_notation_little_e",
+      "format_number_in_scientific_notation_big_e",
+      "format_number_in_binary",
+      "format_number_in_octal",
+      "format_number_in_hexadecimal"
+    ]
+  },
+  "strings": {
+    "is_primitive_or_not": {
+      "name": "Is this a built-in type in this language?",
```
geekygirlsarah: Whenever you're ready, remove all the `name` lines as the text needed is in the meta files and isn't pulled from this file.

web/thesauruses/csharp/strings.json
```diff
@@ -0,0 +1,342 @@
...
+      "name": "Maximum length of string",
+      "code": "1073741823"
+    },
+    "clear_string": {
+      "name": "Clear/empty string",
+      "code": ""
+    },
+    "is_empty": {
+      "name": "Is string empty?",
+      "code": "string.IsNullOrEmpty(string);\nstring.IsNullOrWhiteSpace(string);",
```
geekygirlsarah: This can get really confusing as you mentioned above `string` and `String` are data types, so to have them as possible variable names means this line is pretty hard to decipher. Is the first `string` a variable or the type? and the second?

Maybe use something like `stringName.IsNullOrEmpty` or `str.IsNullOrEmpty` or something. (And obviously the correct form of it.)

And double-check that there's no other references to `string` as a variable too.

web/thesauruses/csharp/strings.json
```diff
@@ -0,0 +1,342 @@
...
+      "name": "Split string by locating all substrings",
+      "code": "str.Split(substring);"
+    },
+    "merge_lists_into_string": {
+      "name": "Merge a list of strings into one string",
+      "code": "string.Join('', stringList);"
+    },
+    "encode_html_entities": {
+      "name": "Encode HTML entities in a string (ex: ™ to &trade;)",
+      "code": "HttpUtility.HtmlEncode(str);"
```
geekygirlsarah: Do you need to import anything to use `HttpUtility`? If so, maybe include that too. (ex: `import Network.HttpUtility;\n\nHttpUtility.HtmlEncode(str);` to show it's needed here. I also just made up the import so I'm not sure if it's right, lol)

geekygirlsarah:
I know this is a draft but figured I'd do a preliminary look over since it popped into my notifications. 

Thanks for working on this issue!

web/thesauruses/csharp/strings.json
blockingHD: Can't have a file about strings without some random references! :)

web/thesauruses/csharp/strings.json
blockingHD: Whoops missed these thanks for spotting them.

web/thesauruses/csharp/strings.json
blockingHD: Using System.Web

blockingHD:
@geekygirlsarah I have gone through and fixed up all the comments and added in the extra bits I was not sure about. Currently I think the Regex functions should be kept and can be reviewed once the meta file is updated against #369 is fixed.

Having said that this also only includes strings and not chars so I think there are some bits missing which might be valuable to include in some other way. But not all languages have chars.

geekygirlsarah:
> 
> 
> @geekygirlsarah I have gone through and fixed up all the comments and added in the extra bits I was not sure about. Currently I think the Regex functions should be kept and can be reviewed once the meta file is updated against #369 is fixed.
> 
> Having said that this also only includes strings and not chars so I think there are some bits missing which might be valuable to include in some other way. But not all languages have chars.

If you have suggestions, feel free to drop them in a new issue or we can keep chatting on this.

But this looks good! I'll merge it in.

Thank you for another PR!

PR #382, opened 2021-10-20
Support array code blocks

### LINKED ISSUE #372: Allow parsing code blocks as array
I'm a newbie dev so I might be wrong, but I think the array/ list approach would result in better readability - presently it does not work I tested it.

_Originally posted by @shy-tan in https://github.com/codethesaurus/codethesaur.us/pull/354#discussion_r729988148_

--- 

Note that this is an easier Django change, but will cause a possible need to alter _every_ JSON file if it doesn't allow single lines _and_ code lines.

RuggMatt:

Closes #372 

## What changed and why?

Allow parsing of JSON files with code blocks defined as array 

e,g, Instead of 
```json
{
  "code": "very long concept\nseveral lines are hard to read \nlets make it easy"
}
```
becomes
```json
{
  "code": [
    "very long concept", 
    "several lines are hard to read", 
    "lets make it easy"
  ]
}
```

Both options are still supported, though I updated the templates to an array containing a single string `[""]` to hint that arrays are supported. Perhaps the docs should also be updated to reflect this

## Checklist

### DIFF
--- a/web/Language.py
+++ b/web/Language.py
@@ -89,7 +89,10 @@ class Language:
         :param concept_key: ID for the concept
         :return: the string containing the concept's code
         """
-        return self.concept(concept_key).get("code")
+        code =  self.concept(concept_key).get("code")
+        if isinstance(code, list):
+            code = "\n".join(code)
+        return code
 
     def concept_comment(self, concept_key):
         """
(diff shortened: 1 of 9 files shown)


geekygirlsarah:
I think when I made the issue, I was originally thinking it would be used for if there was more than one way to do something in code. (The original example linked in the issue was that there's `pow()` and `checked_pow()` in Rust). I was originally thinking this might show them with different background colors or something to differentiate them. But you took it a different direction with one line per array element. This is my fault for not clarifying better.

I may go ahead and merge it in, though I think I want to think through if there isn't a good way to do both. And with so many PRs open, I think it might be ok to pull this in but I want to double-check.

I'll mark it approved/accepted anyway so you get Hacktoberfest credit for now!

geekygirlsarah (approved):

geekygirlsarah:
I think I'll go ahead and merge this now. 

Thanks again @mattschlosser !

PR #389, opened 2021-10-26
IO Langage Concept

### LINKED ISSUE #375: [Concept Request] Basic Input/Output
# Description
I'm not sure if this is within the scope of Code Thesaurus (since this deals more with standard library stuff) but it might be helpful to include.

There are many ways to do input/output handling, but some languages such as C++ provide basic text-based input/output in their standard libraries.

The exact details vary a lot (e.g. reading in a number is easier in C++ than in other languages) - so including related things like string/int conversions might be helpful.

Language Examples
- C++: <iostream> and <sstream>
- C#: System.Console
- Java: Scanner, System.out.println

Function Examples
- Read a string
- Read a line
- Read an integer
- Read a float
- Format strings (if not already covered in Strings)
- Convert integer to string (if not already covered in Strings)
- Convert string to integer (if not already covered in Strings)
- Write a string
- Write a line
- Write an integer
- Write a float

# Requirements

# Additional Notes

blockingHD:

Resolves #375 

## What changed and why?

Added IO meta file

## Checklist

Just outlining currently I know the names need to be variables :)


web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,57 @@
...
+            "Close connection",
+            "Execute query single result set",
+            "Execute query multiple result sets",
+            "Execute non query (update, delete)"
+        ],
+        "Streams": [
+
+        ]
+    },
+    "functions": {}
```
geekygirlsarah: You'll need to replace `functions` with `io`

geekygirlsarah:
It's a start!

geekygirlsarah:
Hey @blockingHD! No rush on this, but figured I'd remind you Hacktoberfest is ending, so if you wanted to finish it by then, you have about 1-2 days left, depending on where you live. If you can get most of it done, I'd be happy to mark it `hacktoberfest-approved` as we still work through minor changes/discussions.

blockingHD:
Not going to get a chance to finish it before the end of hacktoberfest Unfortunately. However will aim to have a reviewable PR by the end of the week.

geekygirlsarah:
That's understandable, but a bummer about Hacktoberfest. The hardest things I've had to do on CT are the meta language files, and they usually take me a few hours each, and I know the project really well, so I imagine they'd take someone else a bit more time. So take your time! I may ping periodically if I don't see anything but it's mostly to just check state, not to be impatient.

blockingHD:
So to many hours later (I didn't quite expect it despite being warned 😅) I think I have everything in here I can think of in the right format? 

Tested this locally and it doesn't appear in the concepts list I am guessing as no language has any info on it yet? If not what do I need to do there.

Lastly I was considering including streams however I do not know enough about them from other languages to be able to include them.

geekygirlsarah:
Haha, I _**did**_ warn you... 😁 

For testing, what I usually do is the exact same thing you'd use to implement it: Copy the _meta file to one of the language directories, then go load up the cheat sheet/reference page with that language and that new concept. If it shows up like you expect, it works. (Just don't commit/push that language copy.)

I can review this later, which I guess can give you the option to test if you want.

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
...
+            "code": ""
+        },
+        "write_line_with_new_line": {
+            "name": "Write a line to the console with a return charactor",
+            "code": ""
+        },
+        "read_line": {
+            "name": "Read a string of user input from the console",
+            "code": ""
+        },
```
geekygirlsarah: You call this "read_line" but say "read a string...", should it maybe be "read a line of..."?

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
...
+        "clear_console_cutput": {
+            "name": "Clear the console output",
+            "code": ""
+        },
+        "change_console_color": {
+            "name": "Change the color of the console",
+            "code": ""
+        },
```
geekygirlsarah: Are you thinking background color, text color, or both? If both, I could see this being different prompts. If one or the other, maybe clarify?

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
...
+        "list_directory": {
+            "name": "List the items in the current directory",
+            "code": ""
+        },
```
geekygirlsarah: Maybe word this like "List of the items..." so it sounds more like it's giving an array/list/vector/something instead of printing out on screen.

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
+{
+    "meta": {
+        "language": "language_id",
+        "language_version": "version.number",
+        "language_name": "Human-Friendly Language Name"
+    },
+    "categories": {
```
geekygirlsarah, line 7: I'm kind of wondering if each section should have like a statement to include. Like C++ has `#include <iostream>`, `#include <fstream>`, etc. to import that functionality. I know Java has it, Python has imports for like networking stuff, PHP has it all built-in though. So this might be a thing we should include.

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
...
+            "code": ""
+        },
+        "send_a_get_request": {
+            "name": "Send a HTTP GET request",
+            "code": ""
+        },
+        "send_a_post_request": {
+            "name": "Send a HTTP POST request",
+            "code": ""
+        },
```
geekygirlsarah: GET/POST are the most common, should we add PATCH/DELETE/whatever the other two are I don't remember off the top of my head?

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
...
+        "add_headers": {
+            "name": "Add headers to a HTTP request",
+            "code": ""
+        },
+        "add_query_params": {
+            "name": "Add query parameters to a HTTP request",
+            "code": ""
+        },
+        "inbuilt_url_class": {
+            "name": "The inbuilt URL class",
```
geekygirlsarah, line 229: What do you mean by this? Do you have an example in some language to what you're referring to?

web/thesauruses/_meta/io.json
```diff
@@ -0,0 +1,243 @@
...
+            "deserialize_json",
+            "serialize_json_with_options",
+            "deserialize_json_with_options"
+        ],
+        "XML": [
+            "serialize_xml",
+            "deserialize_xml",
+            "serialize_xml_with_options",
+            "deserialize_xml_with_options"
+        ]
```
geekygirlsarah: This is probably a good idea! Maybe add CSV too? (CSV tends to have a ton of options too, but maybe we can keep it simpler?)

geekygirlsarah:
Let me know what you think of all of the above thoughts!

web/thesauruses/_meta/io.json
blockingHD: This is kind of tricky as alot of languages have a single 'Make http request' function which you pass the HTTP method into. Hence I went for Get and Post although I had originally also put a make http request section in which would cover the rest. Happy to do the 9 individually, just the most common plus a make http request, or just a make http request I'll let you have the preference :)

web/thesauruses/_meta/io.json
blockingHD: In C# under system there is a Uri class, in Java it is URI under java.net, Typescript and I think now JS have a URI class implementation as well. All used as a standard way to handle URL/Is in a standard way and to validate them against respective standards.

web/thesauruses/_meta/io.json
blockingHD: I did think about CSV however unlike JSON and XML there doesn't tend to be a inbuild language or framework feature to parse them at least in my experience. Again happy to add this in if you think it is valuable however.

web/thesauruses/_meta/io.json
blockingHD: Added to everything but console as that is normally included in the base library for many languages.

web/thesauruses/_meta/io.json
geekygirlsarah: Hmm... I guess for comparison's sake, it might make sense to have "make an HTTP request" and then the specific request types. More annoying to type out individually, but probably the most accurate.

web/thesauruses/_meta/io.json
geekygirlsarah: Ah, ok, this makes sense now. Thanks!

web/thesauruses/_meta/io.json
geekygirlsarah: Python has the `csv` built-in library, but I'm not sure about others off the top of my head. I realize it could be more complicated due to the dozen different ways to set them up, but we could just keep it simple.

geekygirlsarah (changes requested):
Ok, I think I like this and am ready to merge it. But I think between when you started this and now, there's a way to do arrays for code lines now. So you can probably do a quick find and replace for:
`code: ""`
with
`code: [""]`
Do that and I'll merge it!

blockingHD:
I will do this this week. Sorry for the delay!!

geekygirlsarah:
That's alright. I'm just happy someone else decided to take on a concept file!

geekygirlsarah (approved):

geekygirlsarah:
I think I'll go ahead and merge this, but another PR has changed how the meta files work. I'll fix that part up. 

Thank you for working on this and being the first non-Sarah to work on a concept file! I appreciate it!

PR #448, opened 2021-11-06
Load categories from meta file

### LINKED ISSUE #446: Bug: Compare/Reference matching function bases off lang1's categories file
# Expected Behavior

The meta categories should define how the the language pages (compare/reference) are structured, and the one/two languages are matched up to resemble that. 

For example, for functions, there's:
```
Void Functions
   Function that does not return a value and takes no parameters
   ...
Return Value Functions
   Function that returns a value and takes no parameters
   ...
Lambda/Anonymous Functions
   Anonymous function that takes no parameters
   ...
```
represented by:
```
  "categories": {
    "Void Functions": [
      "void_function_no_parameters",
      ...
    ],
    "Return Value Functions": [
      "return_value_function_no_parameters",
      ...
    ],
    "Lambda/Anonymous Functions": [
      "anonymous_function_no_parameters",
      ...
```
within web/thesauruses/_meta/functions.json

# Actual Behavior

For two languages, it loads categories in the first language, the matches the second to resemble it. The meta language's structure doesn't matter.

For one language, it loads categories on the language. The meta language structure doesn't matter.

Why this is bad: If someone wrote a language file (i.e. Python's functions) and left off categories that don't match the meta language's list, then the second one won't display them, whether they exist or not.

1. Modify any language's set of categories
1. Start the server, then compare or reference that language's concept
1. The site shows based on that language, not keeping it the same as the meta language

# Additional Info

This will be in views.py, starting around line 157 for compare, around 257 for reference.

opeco17:

Resolves #446 

## What changed and why?

Currently, the list of categories that is shown on the page depends on each language json file (e.g. `python/functions.json`).
But the list of categories should depend on the meta json file (`_meta/functions.json`).
This PR is to change the source of categories from language json file to meta json file.

Although Subroutines category is not defined in `python/functions.json`, we can see it on the pages.

reference

compare

## Checklist

- [X] (If editing Django) I have added or edited any appropriate unit tests for my changes

### DIFF
--- a/web/views.py
+++ b/web/views.py
@@ -256,13 +256,17 @@ def reference(request):
 
     categories = []
     concepts = []
-    for category_key in lang.categories:
+    
+    all_category_keys = list(meta_structure.categories.keys())
+    all_concept_keys = list(meta_structure.concepts.keys())
+
+    for category_key in all_category_keys:
         categories.append({
             "id": category_key,
             "concepts": meta_structure.categories[category_key]  # meta_lang_categories[category_key]
         })
 
-    for concept_key in lang.concepts:
+    for concept_key in all_concept_keys:
         concepts.append({
             "id": concept_key,
             "name": meta_structure.concepts[concept_key]["name"],


geekygirlsarah:
Dang... was this _that_ simple of a change? I figured there'd be more rewriting!

But hey, this looks great and tests seem to pass, so I'll merge it in.  Thank you so much!

opeco17:
Thank you for reviewing!

PR #457, opened 2021-11-29
Remove categories from the language files

### LINKED ISSUE #208: Remove need for Categories block from language JSON files
# Description
The categories section in all the languages files look exactly like the ones in the respective _meta files (apart from not being up to date). I think it would make the language files easier to write & look cleaner if we just dropped this duplicate information.

# Additional Info
Now with #204 the need to signify concepts that are not implemented for a language is handled in the respective concept blocks.

cafce25:

Resolves #208

## What changed and why?

- Separate templates from the meta files
- Refactor the meta files to include only relevant, non duplicated information
- Remove redundant category information from the language files

## Checklist

To get rid of the categories sections I used jq wich reformatted the json files resulting in some files being switched from tab indenting to spaces.

I think separating _meta and _templates makes both cleaner which is why I included it here. If this is too out of scope simply cherry picking the "Remove categories from lamguage files" commit should suffice.


web/MetaInfo.py
```diff
@@ -8,7 +9,9 @@ def __init__(self):
         Initializes MetaInfo object with meta language information
         :rtype: None
         """
-        with open("web/thesauruses/meta_info.json", 'r') as meta_file:
+        meta_info_file_path = os.path.join(
```
geekygirlsarah, line 12: omg more variables! this thing is already complex!

(though I approve of this change, lol)

web/templates/compare.html
```diff
@@ -32,22 +32,18 @@ <h3 class="text-center">{{ lang2_friendlyname|linebreaksbr }}'s Implementation</
...
-                            <div class="strong">
-                                {{ c.name|default:"Unsupported or Not Needed"|linebreaksbr }}
-                            </div>
-                        </div>
+    {% for concept in category.concepts %}
+        <div class="card-group">
+            <div class="card">
+                <div class="card-body">
+                    <div class="strong">
+                        {{ concept.name|default:"Unsupported or Not Needed"|linebreaksbr }}
```
geekygirlsarah, line 40: Huh, didn't know there was a default here. Cool!

web/templates/reference.html
```diff
@@ -26,21 +26,17 @@ <h3 class="text-center">{{ lang_friendlyname|linebreaksbr }}'s Implementation</h
...
-                    <div class="card-body">
-                        <div class="strong">
-                            {{ c.name|linebreaksbr }}
-                        </div>
+    {% for concept in category.concepts %}
+        <div class="card-group">
+            <div class="card">
+                <div class="card-body">
+                    <div class="strong">
+                        {{ concept.name|linebreaksbr }}
```
geekygirlsarah: Shouldn't this also have a default?

geekygirlsarah:
I'm realizing your PR and it's 110+ file changes is REALLLLLLY going to break my language validation script I've been working on ( #454 )

😬😬😬

web/templates/reference.html
cafce25: I think rather than adding default here we should remove it above, if I'm not mistaken we handle missing concepts in the backend and should never need to have a default here.

cafce25:
If you prefer we could put the _meta/_templates separation which probably brings the most breakage or even this whole PR on hold and revisit it after you're done with your PR. I'm happy either way.

geekygirlsarah:
> If you prefer we could put the _meta/_templates separation which probably brings the most breakage or even this whole PR on hold and revisit it after you're done with your PR. I'm happy either way.

I'd rather help keep changes that other contributors (including you) do and I'll work around it. It's my project, but I'd rather it be easier for others to work with. It'll be fine, it'll probably make some of the validation easier in the long run once we've ironed out the questions.

I still like the _templates idea, I just also wonder if the upkeep of two files that are almost exactly the same is worth it? Also since the `name` lines aren't needed in the language file, but they are handy so you know what the concept IDs mean when you're making it. I haven't quite figured out the best way for doing this. (I did think it'd be cool to have an online editor and just avoid messing with JSON files all together, but that seems so far down the road at this point.)

web/templates/reference.html
geekygirlsarah: Yeah, it should be in the backend too. I thought you were moving it from backend to frontend. Whichever!

cafce25:
I'm totally with you on having a template generator rather than a _templates directory. I've juggled the idea in my head a little and think maybe we can include it in the website.
I'll play around with it some and will check back once I have something working.

cafce25:
I've added the templates onto the error pages that get shown upon hitting a language/structure combination that is not implemented it. 

Code-Thesaurus with template

Code-Thesaurus with template expanded

In addition I've added a `generate_template`-command to `./manage.py`: `./manage.py generate_template java data_types` will now generate the `data_types` concept file for `java` and save the skeleton to the corresponding folder, you can see other options with `./manage.py generate_template -h`

The github link now fills in the filename as well which is why i removed it from the text before that.

Looking forward to your feedback!

web/templates/reference.html
cafce25:
```suggestion
                        {{ concept.name | default:"Unsupported or Not Needed" | linebreaksbr }}
```

geekygirlsarah:
I dropped the ball on this... sorry about that!

I like this template idea more than I thought I would because I think I expected it to work differently. I'd be happy to merge this in, but one change I might ask is to add the `name` lines back in. I realize it's a step back, but those ID names don't make any sense on their own. If I knew the contributor would test them or look them up or view the meta side-by-side with the added file, or even pull open CT locally to look at it, I might not mind, but I think it lacks context and might leave them guessing what it is. But then again, it's that step into adding a thing that's not needed.

I'm open to ideas though... any thoughts?

(Also I made a review app/deployment at https://codethesaurus-pr-457.herokuapp.com , if that's helpful at all.

cafce25:
All right, added in all the `name`s, while I was at it I updated them in the thesaurus files. The template generator generates the `name` fields as well.

geekygirlsarah (approved):

geekygirlsarah:
Alright, I think it's good to go. I'll merge it!

Thanks for all the work on this! I hope you have a lovely holiday season!

PR #328, opened 2021-10-06
Add database to track lookups

geekygirlsarah:
## What GitHub issue does this PR apply to?

Resolves #321 

## What changed and why?

Added database back into the app so that it can track what is looked up in the app. Also added migration capabilities to the Heroku Procfile.

## (If editing Django app) Please add screenshots

Image of site visit table:
image

Image of site lookup table:
image

## Checklist

- [x] I claimed any associated issue(s) and they are not someone else's
- [x] I have looked at documentation to ensure I made any revisions correctly
- [x] I tested my changes locally to ensure they work
- [ ] (If editing Django) I have added or edited any appropriate unit tests for my changes

PR #459, opened 2021-12-01
Fix linting errors and merge the different models

cafce25:
## What changed and why?
- Fix most linting errors.
- Merge the models `Language`, `MetaInfo` and `MetaStructure` into one file `models.py`
- Replace String appending with using f-Strings instead (Better readability, less code, better performance) where possible.

## Checklist

- [X] I claimed any associated issue(s) and they are not someone else's (there are none)
- [X] (If editing Django) I have added or edited any appropriate unit tests for my changes

The commit history is a little verbose cause I wanted to keep the history for Language, MetaInfo and MetaStructure.


geekygirlsarah:
I think when it comes to MVC-type sites, the models directly correspond to data stored in databases. I know that in a way Language and the other objects essentially are kind of those things except from data files instead of databases, but Django's models have ORM save/get methods attached to them. (I started making a couple of those in #328, but it's not merged in yet as there's some config changes I need to make and test). 

Also merging those models in seems to be extraneous compared to what the PR _sounds_ like it's supposed to do, which is just fix linting errors. 

I'm open to hearing the reasoning behind this though.

cafce25:
I merged the files into models for a few reasons:
- to me models are the logic that connect the data to the business logic which `MetaInfo`, `MetaStructure` and `Language` do
- `pylint` picked up on the models file basically being empty, so at first I deleted it
- but then it also remarked about the file names `MetaInfo.py`, `MetaStructure.py` and `Language.py` which according to it should be snake cased
- imports become cleaner with them residing in the same module and grouping them makes sense to me

geekygirlsarah:
> I merged the files into models for a few reasons:
> 

Hmm.. alright, I'll allow this @cafce25. Go on...

(It's also a draft. Not sure if you're done tinkering yet...)

cafce25:
Not done yet, sort of waiting on #457 since that fixes some lint's / will conflict with this.

geekygirlsarah (approved):

geekygirlsarah:
Alright, let's do it. Merge all the models!

Issue #188, opened 2021-04-18
Have CT link to documentation on each item


geekygirlsarah:
# Description

Each file (maybe each ID?) should link to the appropriate place of documentation for that language.

# Requirements

Add it to each file, and discuss with @geekygirlsarah on if it should be per item, per category, or per file.

saracope:
@geekygirlsarah Mind clarifying this one for me? Would this be additional meta info?

geekygirlsarah:
Yeah. So instead of something like:
```
"void_function_no_parameters": {
    "code": ["void FunctionName()\n{\n\t//code\n}"]
    "comment": "Can also be static void"
}
```
It'd add in an additional "documentation" line or something:
```
"void_function_no_parameters": {
    "code": ["void FunctionName()\n{\n\t//code\n}"],
    "comment": "Can also be static void",
    "documentation": "https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions"
}
```

or whatever. I'm open to what it's called or how it's implemented. But my thought was if someone wanted to go read more on something, they could click on a link to go there.

PR #467, opened 2022-02-13
Add version handling to Language

### LINKED ISSUE #263: Rename thesaurus folders to reflect language version numbers
# Description

In `web/thesauruses/`, all the folders just consist of the name of the language. This issue is to allow for things like `python2` and `python3` to exist to allow for comparing language versions too.

# Requirements

1. The easy part: rename the folders to their latest _major_ versions. So `python` to `python3` is fine. `python` to `python391` or `python3.9.1` is not.
2. Update `web/thesauruses/meta_info.json` to reflect the new folder names.
3. The hard part: The syntax highlighting is based on the folder name. This need to be updated to have an additional component to know what language it's in. This could be done in a couple of ways:
    a. Change the language section of `meta_info.json` to be a key (ID) with a value that's an object instead of a string. Add in the language syntax ID in there. This is probably the more "robust" or "correct" solution.
    b. Add the syntax highlighting ID into every _meta language file. This will be simpler to build, but also must have _every_ language file also altered to reflect this (which means changing probably at least 30 language files to also have the same changes).
4. Ensure the Django code reflects whatever change you made.
5. Test it! (It's a larger change that may break some of the built in tests.)

# Additional Notes

This will likely be a big pull request, so it may take longer to review and test. Please be patient! If you're doing this for Hacktoberfest, I'll ensure it gets tagged appropriately and before the end of the month.

cafce25:

Resolves #263

## What changed and why?

Move the structure files into the corresponding version sub folders
Add structure version merging logic to `models.Language`

## Checklist

The version string parser used is `packaging.version.parse` which i checked works with the languages currently in codethesaur.us.

The merging algorithm I implemented considers all versions prior to the one specified i.e. for Python 3.7 it will check any version before that like `3.6.x`, `3.6`, `3.5` ... but also different major versions like `2.x` and will take any bare `python/<structure>.json` as base case. I believe this to be sensible since it's all the same language after all but maybe we want to stop at the major version (`3`) instead?
When providing no version info at all it will take all `python/**/<structure>.json` files into account.


web/thesauruses/cpp/17/data_types.json
```diff
@@ -1,7 +1,7 @@
 {
   "meta": {
     "language": "cpp",
-    "language_version": "C++17",
+    "language_version": "17",
```
geekygirlsarah: The versions are actually called "C++10", "C++14", "C++17", etc. it's based on the year and they actually do include the `C++` in them.

geekygirlsarah:
I appreciate what you started with this, though I think maybe this is sort of bypassing part of what I wanted to set up and making some other parts a bit harder to maintain.

I really only want to track major versions. Compare Python 2 vs Python 3. Ruby 2 vs Ruby 3. Things like that. The goal is to see how the major versions can evolve, and do things to help people upgrade their code (like old Python 2 projects still exist because people find it's really hard to migrate them). Comparing Python 3.9 and 3.10 is a little too nit-picky, and to get into 3.9.1 vs 3.9.2 is way too much work and mostly going to result in multiple files without changes. It's going to make the scale of this thing grow at a rate no one can keep up with, and to make minor tweaks for updating is going to involve moving files and folders around and it seems a bit more work than I'd like to have. 

It's kind of looking like all the versions are getting bundled up together (like the two C++ versions) merge into one language-picker drop-down on the home page). That's also against what I was hoping the original issue would create which is separate entries for those major versions. (Though will admit I haven't finished looking through this enough to tell if I put the same concepts in other version directories if they'd show up differently yet.)
image

cafce25:
> <details><summary>I really only want to track major versions.</summary> Compare Python 2 vs Python 3. Ruby 2 vs Ruby 3. Things like that. The goal is to see how the major versions can evolve, and do things to help people upgrade their code (like old Python 2 projects still exist because people find it's really hard to migrate them). Comparing Python 3.9 and 3.10 is a little too nit-picky, and to get into 3.9.1 vs 3.9.2</details>

I think the support for at least minor versions is required because of things like Rust format strings etc which only got added later. For example:
```rust
let v = "abc";
println!("{v}");
```
is valid Rust 1.59 but not 1.58 (not 100% sure on the exact version numbers),
Pythons `async/await` syntax also hit past 3.0 iirc.

> <details><summary>is way too much work</summary> and mostly going to result in multiple files without changes. It's going to make the scale of this thing grow at a rate no one can keep up with, and to make minor tweaks for updating is going to involve moving files and folders around and it seems a bit more work than I'd like to have.</details>

but I think I haven't communicated the algorithm detailed enough.
Concepts that are the same across all of Python 3 would go into `python/3/<structure>.json`, all later versions will pick those concepts up automatically, you'd only have to create another file if you have concepts that only apply to versions after the major version release.

I have moved all the files into the directory of the version specified in the file, maybe moving them all to the `<language>/<major>` directory would be more appropriate to start out, should I do that?

> <details><summary>It's kind of looking like all the versions are getting bundled up together</summary> (like the two C++ versions) merge into one language-picker drop-down on the home page). That's also against what I was hoping the original issue would create which is separate entries for those major versions. (Though will admit I haven't finished looking through this enough to tell if I put the same concepts in other version directories if they'd show up differently yet.)</details>
Yess, that's the thing I wanted to add still 😅, slipped by me, I'll add it.

geekygirlsarah:
>I think the support for at least minor versions is required because of things like Rust format strings etc which only got added later. For example: [...] is valid Rust 1.59 but not 1.58 (not 100% sure on the exact version numbers), Pythons `async/await` syntax also hit past 3.0 iirc.
> 

Hmm... Alright, you have a point. Though it makes me wonder if major versions should get combined together still as people will more likely upgrade 1.58->1.59 without thinking too much breaking where a 1.x->2.x would be a lot more thought behind it. But also maybe adding things like string formatting could be more of a "available since version x.x.xx" kind of a tag added to the concept itself. (Though at that point it seems overkill and they can look at language docs if they want that.) I guess I'm going back and forth between CT being just a thesaurus vs being an all inclusive language reference thing. I've mostly been keeping it as close to a thesaurus as possible and shrugging off that would mean the project would have to maintain tons of meta data about things, which also means tons of upkeep and possible invalid data. (One example would be if a concept doesn't exist in the language but you can write code in 2-3 lines to replicate it. Yeah, it recreates it, but it's not a part of the language, so I've told people to not include those. Because where do we stop at the point of writing code to replicate features of other languages?)

I may ponder on this more.

> but I think I haven't communicated the algorithm detailed enough. Concepts that are the same across all of Python 3 would go into `python/3/<structure>.json`, all later versions will pick those concepts up automatically, you'd only have to create another file if you have concepts that only apply to versions after the major version release.
> 
> I have moved all the files into the directory of the version specified in the file, maybe moving them all to the `<language>/<major>` directory would be more appropriate to start out, should I do that?

I can appreciate this thought, but also as maintainer of this project, I want it to be easy for people to figure out and work with. If one concept starts becoming nested in multiple files, it loses its ease of working with, especially as they're JSON files and there's no easy way to load multiple files, merge them, then edit that result. The trade off is the further a minor/patch version gets from the latest of that minor, it might get less accurate. I'm not sure where that line is that CT should keep yet.  I think I've felt simpler is better though, and hoped that with a `version` key at the top of the concept files would mean if something changed in a minor version, people would update that minor version too and just keep the latest minor for all languages.

> Yess, that's the thing I wanted to add still 😅, slipped by me, I'll add it.

Oops, lol 😁

colindean:
What about adding some metadata to examples that indicate a "since" version, with perhaps a "deprecated" or "removed" version as well? Having a directory structure per-version is exponentially more difficult to manage. Having been thrust into maintaining several Python codebases, in the last year, the official Python docs providing a `since` version is incredibly important because I'm working across projects started in the last few years with 3.5 all the way to 3.10.

geekygirlsarah:
I think @colindean and I think alike...

I also made issue #188 as a way that maybe if someone wanted to dive further into a concept could click on a link and go to the full documentation of a thing. Not sure if this is better than a `since`/`deprecated` version tracker, or if there should be both.

cafce25:
Okay, I'm in the middle of reworking this, but I'm torn between putting every _C++_ or _Javascript_ (or any of the date of publishing based versions) as a major version or treat them all like minor versions since they're all backwards compatible as far as I can tell.

Another thing I'm not sure how to handle is should I throw out the merging code completely or should we keep it around and use it for major versions, I think most concepts won't change across versions and we can reduce the amount of duplication that way but as you said it becomes more complex to find the right file to edit.

geekygirlsarah:
If there's no minor versions (`C++20.1`), then they're major versions.

I _want_ to say toss the merging out because it makes editing the files and knowing what previously exists and what to add in extremely difficult. I think if CT had an editing tool* this would be fine but we don't. It means duplication, but perhaps not as much work? (As in like when a new version comes out, you can copy the previous version to a new folder, then modify what changed, and the old would, in theory, still be right.) I like the idea of the merging, but I think it just makes the idea of knowing where stuff exists and where to add new stuff more complicated.

* This is like a background thought in my mind but I'm trying to do so many other things first.

geekygirlsarah (approved):

geekygirlsarah:
This is good. I have a couple of minor things but can fix those later. You've done enough work on this, so I'll merge it in!

Thank you again for taking on some of the larger issues, but also trying to trying out new things I haven't thought of!

PR #478, opened 2022-04-13
Issue 290 - add Lists/Arrays/Hashes for Python3

### LINKED ISSUE #290: Add Python Lists/Arrays/Hashes
# Description

Lists/Arrays/Hashes for Python aren't presently on Code Thesaurus. This issue is to add that concept to the system.

# Requirements

* Copy `web/thesauruses/_meta/lists.json` to the right Python directory in `web/thesauruses/`
* Edit the lists.json file to accurately reflect the latest major version of Python
* Add in `"not-implemented": "true"` if the concept doesn't exist in the language. You can then leave `"code":` blank.

# Claiming This Issue

# Need Help?

# Thanks!

TomReilly1:

Resolves #290 

## What changed and why?

Created the 'lists.json' file to provide concepts for the Lists/Arrays/Hashes structure for Python 3.

## Checklist

### DIFF
--- a/web/thesauruses/python/3/lists.json
+++ b/web/thesauruses/python/3/lists.json
@@ -95,9 +95,6 @@
       "comment": "First line creates an empty set. Second line creates a set with one value. Third line creates a set with multiple values"
     },
     "unordered_mutable_list_start_number": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "unordered_mutable_list_can_be_appended": {
@@ -112,21 +109,12 @@
       "comment": "Because a set is unordered, the index cannot be specified"
     },
     "access_element_in_unordered_mutable_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "insert_into_beginning_of_unordered_mutable_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "insert_into_end_of_unordered_mutable_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "insert_into_middle_of_unordered_mutable_list": {
@@ -136,9 +124,6 @@
       "comment": "Because a set is unordered, the index cannot be specified"
     },
     "erase_element_at_beginning_of_unordered_mutable_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "erase_element_at_end_of_unordered_mutable_list": {
@@ -153,9 +138,6 @@
       ]
     },
     "swap_elements_in_unordered_mutable_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "delete_unordered_mutable_list": {
@@ -203,15 +185,9 @@
       "comment": "The frozenset() function can be used on any iterable object. However, to get an unordered immutable list, it should be used on a set."
     },
     "unordered_immutable_list_start_number": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "access_element_in_unordered_immutable_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "delete_unordered_immutable_list": {
@@ -226,57 +202,30 @@
       "not-implemented": "true"
     },
     "create_a_mutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "insert_element_to_mutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "erase_element_from_mutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "delete_mutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "name_of_immutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "create_a_immutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "insert_element_to_immutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "erase_element_from_immutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "delete_immutable_hashed_list": {
-      "code": [
-        ""
-      ],
       "not-implemented": "true"
     },
     "create_a_mutable_set": {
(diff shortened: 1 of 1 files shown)


web/thesauruses/python/3/lists.json
```diff
@@ -0,0 +1,492 @@
...
+    "swap_elements_in_ordered_mutable_list": {
+      "code": [
+        "list_name[index1], list_name[index2] = list_name[index2], list_name[index1]"
+      ]
+    },
+    "delete_ordered_mutable_list": {
+      "code": [
+        "del list_name"
+      ]
+    },
```
cafce25, line 83: These (L 9-83) should be in `*unordered_list*`, i don't think python provides ordered lists.
Not sure if it's me or sth but github seems to delete my line number annotations so I'm putting them explicitly.

web/thesauruses/python/3/lists.json
```diff
@@ -0,0 +1,492 @@
...
+    "swap_key_and_value_in_mutable_set": {
+      "code": [
+        "dict_name = dict([(value, key) for key, value in dict_name.items()])"
+      ]
+    },
+    "delete_mutable_set": {
+      "code": [
+        "del dict_name"
+      ]
+    },
```
cafce25, line 271: Python `dict`s are what codethesaurus calls `hashed lists` and should be in the `*mutable_hashed_list*` concepts.
L 282-322

web/thesauruses/python/3/lists.json
```diff
@@ -0,0 +1,492 @@
...
+      "code": [
+        ""
+      ],
+      "not-implemented": "true"
+    },
+    "delete_unordered_mutable_list": {
+      "code": [
+        "del set_name"
+      ]
+    },
```
cafce25: Above (L 9-83) concepts should go here, these (L 84 - 165) are set operations and belong in the `*set*` definitions below.

web/thesauruses/python/3/lists.json
```diff
@@ -0,0 +1,492 @@
...
+    "access_element_in_ordered_immutable_list": {
+      "code": [
+        "tuple_name[index]"
+      ]
+    },
+    "delete_ordered_immutable_list": {
+      "code": [
+        "del tuple_name"
+      ]
+    },
```
cafce25, line 175: I'm not sure about putting tuples as immutable lists, sure in python you can treat them as such but in my mind a tuple is fundamentally different from a list.
To me lists are collection of an unspecified amount of same/similar elements,
while tuples are a often heterogenous collection of a fixed amount of items.

web/thesauruses/python/3/lists.json
```diff
@@ -0,0 +1,492 @@
...
+        ""
+      ],
+      "not-implemented": "true"
+    },
+    "find_element_by_value": {
+      "code": [
+        "value in list_name"
+      ],
+      "comment": "returns a True/False value"
+    },
```
cafce25: I believe this should be `list_name.index(value)`

web/thesauruses/python/3/lists.json
```diff
@@ -0,0 +1,492 @@
...
+        "set_name = {}",
+        "set_name = {value}",
+        "set_name = {value1, value2, ...}"
+      ],
+      "comment": "First line creates an empty set. Second line creates a set with one value. Third line creates a set with multiple values"
+    },
+    "unordered_mutable_list_start_number": {
+      "code": [
+        ""
+      ],
```
cafce25: These empty code blocks for unimplemented concepts can be removed.

cafce25 (changes requested):
Hey, thanks for taking the time to write this up, very much appreciated!

you mixed up some of the concept <-> sample mappings, (which isn't hard cause all sorts of languages call the different concepts all sort of things)
in summary I think:
- everything under \*ordered_list* should be in \*unordered_list*
- everything under \*unordered_list* should be in \*set*
- everything under \*set* should be in \*hashed_list*
- all the things in \*ordered_list* should be declared unimplemented.
- tuple might be a candidate for new concepts instead of trying to fit it into immutable ordered lists (i can think of a couple of languages that have them & immutable lists).

Other than that I think you did a great job.

web/thesauruses/python/3/lists.json
TomReilly1: I think the biggest difference between a Python list and a Python tuple is mutability, which naturally leads to the former being dynamic and the latter being fixed. In terms of of heterogeneous vs. homogeneous, both Python lists and Python tuples are capable of storing either type of collections, even if one makes more sense for the other.

I am not too familiar with the differences between tuples and immutable lists from other languages, but if the main difference is what data each contains (heterogeneous vs. homogeneous), I think Python tuples could be classified as both a tuple and immutable list.

web/thesauruses/python/3/lists.json
TomReilly1: Python lists are ordered in a sense that its elements have a specific order, unless CT's definition of ordered is different than that.

web/thesauruses/python/3/lists.json
TomReilly1: Will do

web/thesauruses/python/3/lists.json
TomReilly1: Could have both since they both find by value. Would comment that one returns a boolean and the other returns an index/integer

web/thesauruses/python/3/lists.json
cafce25: On rereading this seems I mixed it up here, of course `*set*` is the right place for these.

web/thesauruses/python/3/lists.json
cafce25: You're right. CT uses it inconsistently at the moment and I've created #481. Let's decide one way or the other over in the issue, I think leave it here then and get merged in the meantime.

web/thesauruses/python/3/lists.json
cafce25: Same as above, leave it here, decide later.

web/thesauruses/python/3/lists.json
cafce25: Hmm yes, I thought of it as either or, but it fit's both so why not add it in both places (when we add tuples)!

web/thesauruses/python/3/lists.json
geekygirlsarah: The definition of "ordered" that I was thinking of when I added lists corresponds to Python's lists being ordered lists.

web/thesauruses/python/3/lists.json
geekygirlsarah: This sounds resolved?

geekygirlsarah (changes requested):
I think this is pretty good. Sounds like cafce25 found a couple of things to fix, but I can probably merge in after that!

TomReilly1:
OK I think we are good now. Removed the empty code blocks and added the method for finding an element by value.
I left the lists, sets, and tuples as they were.

geekygirlsarah (approved):

geekygirlsarah:
Thank you for working on this and for all the changes! I'll merge it in.

PR #479, opened 2022-04-15
add handle to support 100+ contributors on footer

### LINKED ISSUE #465: Contributor count on bottom of pages will max out at 100
# Expected Behavior

The contributor count on the bottom of every page should display the _actual_ count of contributors.

# Actual Behavior

The GitHub API limits calls to 100 contributors per "page", and only one page of results is fetched. This number will no longer be correct after we hit 101 contributors.

# Additional Info

In `web/static/js/contributors.js` is where the code to fetch the number. This will need to be rewritten to account for multiple pages of results if over 100 people.

splashley:

Resolves #465

## What changed and why?

The GitHub API limits calls to 100 contributors per "page", and only one page of results was being fetched. Updated function to make additional API calls when there are > 100 contributors.

## Checklist

### DIFF
--- a/web/static/js/contributors.js
+++ b/web/static/js/contributors.js
@@ -1,5 +1,22 @@
-$(document).ready(function(){
-  jQuery.get('https://api.github.com/repos/codethesaurus/codethesaur.us/contributors?accept=application/vnd.github.v3+json&per_page=100&anon=1', function(data, status){
-    $("#contributors").text(data.length)
-  })
-})
+$(document).ready(function () {
+  let i = 1;
+  let length = 0;
+  const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+  const timer = async () => {
+    await sleep(1000);
+    jQuery
+      .get(
+        `https://api.github.com/repos/codethesaurus/codethesaur.us/contributors?accept=application/vnd.github.v3+json&per_page=100&page=${i}&anon=1`
+      )
+      .then((data) => {
+        length += data.length;
+        i++;
+        if (data.length === 0) {
+          $("#contributors").text(length);
+        } else {
+          timer();
+        }
+      });
+  };
+  timer();
+});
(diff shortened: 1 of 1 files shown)


web/static/js/contributors.js
```diff
@@ -1,5 +1,22 @@
...
-    $("#contributors").text(data.length)
-  })
-})
\ No newline at end of file
+$(document).ready(function () {
+  let i = 1;
+  let length = 0;
+  const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+  const timer = async () => {
+    await sleep(1000);
```
cafce25: Waiting before the request seems to me like wasting time can we move this down to before L17 where we call timer again? Maybe something like `setTimeout(timer, 1000)` in L17 ?
Also can we maybe get away with a shorter delay between calls 1s seems long to me but it might be necessary for the github api?

cafce25 (commented):
Yeah! Now the contributors cane come in hoardes!!

We might wanna use the link headers rel="last" link (`link | …,  <https://api.github.com/repositories/106052676/contributors?page=3>;  rel="last"`) to get the contributor count in 2 api-requests by `nContributors * (pages-1) + contributors_last_page` but this should be plenty for the time being.

web/static/js/contributors.js
splashley: ~~Waiting before the request seems to me like wasting time can we move this down to before L17 where we call timer again? Maybe something like setTimeout(timer, 1000) in L17 ?~~
Done.

_Also can we maybe get away with a shorter delay between calls 1s seems long to me but it might be necessary for the github api?_
I cut it down to 500ms. I believe this will be sufficient for now.

geekygirlsarah:
I think this will work for now and can be tweaked later.

Thank you @splashley (and @cafce25 for reviewing)! I'll merge it in.

geekygirlsarah (approved):

PR #485, opened 2022-04-25
removing jquery dependency

### LINKED ISSUE #482: Remove JQuery dependency
# Description

Bootstrap has been upgraded to version 5 which is not dependent on JQuery. We should stop using JQuery in base.html, and update contributors.js to use either fetch, or if we need IE 11 support, then XMLHttpRequest will do.

# Requirements

We need to know whether IE 11 is supported for the project, and if so, that'll decide on what implementation we use for contributors.js.

# Additional Notes
This will bring benefits to FCP and LCP (lighthouse), and reduce the amount of resources needed for users to download.

barretojs:

Resolves #482

## What changed and why?

- Removed Jquery import from `base.html`
- changed `contributors.js` to use XMLHttpRequest to get the request from github.
- added error handling in case the request gets messed up somehow, by setting contributors to 0 (only if the data is not available)
- increased the `per_page` attribute from the url, because the repo is nearing 100 contributors!

## Checklist

### DIFF
--- a/web/static/js/contributors.js
+++ b/web/static/js/contributors.js
@@ -10,7 +10,7 @@ document.addEventListener("DOMContentLoaded", function () {
 
   contributorsRequest.onload = function () {
     if (contributorsRequest.status != 200) {
-      document.querySelector("#contributors").innerHTML = 0;
+      document.querySelector("#contributors").innerHTML = "multiple";
       return;
     }
 
@@ -19,11 +19,11 @@ document.addEventListener("DOMContentLoaded", function () {
 
       document.querySelector("#contributors").innerHTML = response.length;
     } catch {
-      document.querySelector("#contributors").innerHTML = 0;
+      document.querySelector("#contributors").innerHTML = "multiple";
     }
   };
 
   contributorsRequest.onerror = function () {
-    document.querySelector("#contributors").innerHTML = 0;
+    document.querySelector("#contributors").innerHTML = "multiple";
   };
 });
--- a/web/static/js/contributors.js
+++ b/web/static/js/contributors.js
@@ -23,11 +23,12 @@ document.addEventListener("DOMContentLoaded", function () {
       try {
         var response = JSON.parse(contributorsRequest.response);
 
-        if (!response.length) {
+        numberOfContributors += response.length;
+
+        if (response.length < 100) {
           document.querySelector("#contributors").innerHTML =
             numberOfContributors;
         } else {
-          numberOfContributors += response.length;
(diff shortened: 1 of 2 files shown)


web/static/js/contributors.js
```diff
@@ -1,5 +1,29 @@
...
+  contributorsRequest.open(
+    "GET",
+    "https://api.github.com/repos/codethesaurus/codethesaur.us/contributors?accept=application/vd.github.v3+json&per_page=200&anon=1"
+  );
+
+  contributorsRequest.send();
+
+  contributorsRequest.onload = function () {
+    if (contributorsRequest.status != 200) {
+      document.querySelector("#contributors").innerHTML = 0;
```
NathanBeddoeWebDev: Hmm, wonder if we could show a better error. Maybe hide the overarching element altogether? 0 contributors will look kind of weird

geekygirlsarah: Presently if the call never happens, I think it just shows "and contributors." Maybe it could get changed to "and multiple contributors" or something like that?

image

geekygirlsarah (changes requested):
I think this idea will work, just change the error.

It looks like another PR and this one conflicted, so this might need to integrate those changes. (It's the "GitHub API only gets 100 contributors at a time" issue. The other PR fixed it, this code doesn't include that fix.)

web/static/js/contributors.js
barretojs: i changed it to multiple contributors, it's better than zero indeed.

barretojs:
i merged the previous pagination pr and added a few tweaks. tested with repositories with more contributors and worked fine.

barretojs:
hey, is there any more problems with this pr for me to fix?

geekygirlsarah:
This is a side project of mine, so patience is always appreciated as I still have a day job and other things going on too.

But yeah, I think it looks good. I'll merge it in! Thank you for working on this!

barretojs:
sorry. it was not my intention to sound impatient or pressuring the merge.
we'll keep an eye on if i can contribute to other stuff too. thanks for the cool project.

geekygirlsarah:
> sorry. it was not my intention to sound impatient or pressuring the merge.
Sorry if I misread or misunderstood, too!

>we'll keep an eye on if i can contribute to other stuff too. thanks for the cool project.

Thanks! I try to keep issues up to date, and if you ever see problems or have ideas, feel free to add an issue for them too! This project really only works by people continuing to contribute, so I appreciate everyone's help.

PR #486, opened 2022-05-09
Fix links to contribute

### LINKED ISSUE #484: Add/Correct links on bottom of compare/reference pages no longer work
# Expected Behavior

The links at the bottom of reference and compare pages should go to the correct edit page in CT's Github repo.

image

# Actual Behavior

The links go to a 404 page instead.

1. Compare things or look up a reference page (ex: [functions in C# and Python](https://codethesaur.us/compare/?concept=functions&lang1=csharp%3B9&lang2=python%3B3)
1. Click on one of the "Want to add or correct information for..." links
1. Github shows you a 404 page.

# Additional Info

When language versions was added, these links broke. The version number will need to be added into the URL. The templates to edit the URLs in are in `web/templates/` directory but the version number for the language(s) are in `web/views.py`. You'll probably need to pass them to the template somehow.

_Ideally_ we should add tests to ensure those URLs don't break in the future, but i"m not sure how you can actually test that (without maybe pinging GitHub to see if the page returns a 200 vs 404) so this is optional. Please do it if you can figure out how!

teallarson:
https://github.com/codethesaurus/codethesaur.us/issues/484

Resolves #484

## What changed and why?

Adds version numbers to language data returned from `views.py` and adds them to the html templates.

Here's a link that was previously broken 🎉:
https://user-images.githubusercontent.com/63751206/167378875-fe056d29-15da-48fd-bcb5-2769762d745e.mov

## Checklist

This would be awesome to test... the two ways to do this that I can think of are (a) pinging the URLs to see if the link actually gets a 200 or (b) read the file paths in `/thesauruses/` and see that they match the patterns in the generated URLs. Both feel overkill... if you think of a way (or if either of those seem worth it) I'm happy to tinker!

### DIFF
--- a/web/templates/compare.html
+++ b/web/templates/compare.html
@@ -54,14 +54,14 @@
                 </div>
                 <div class="card" id="start-of-row">
                     <div class="card-body">
-                        <a href="https://github.com/codethesaurus/codethesaur.us/edit/main/web/thesauruses/{{ lang1 }}/{{ concept }}.json">
+                        <a href="https://github.com/codethesaurus/codethesaur.us/edit/main/web/thesauruses/{{ lang1 }}/{{ version1 }}/{{ concept }}.json">
                             Want to add or correct information for {{ lang1_friendlyname }}?
                         </a>
                     </div>
                 </div>
                 <div class="card">
                     <div class="card-body">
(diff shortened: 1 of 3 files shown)


geekygirlsarah:
Looks good to me!

And I have no clue how to make tests for this either without pinging Github (which tests shouldn't do) or mocking something but Github changing anything the tests would still pass. So for now we can leave it be.

Thanks! I'll merge it.

PR #555, opened 2022-10-04
Strings in Ruby

### LINKED ISSUE #426: [Ruby] Add strings
# Description

Strings for Ruby aren't presently on Code Thesaurus. This issue is to add that concept to the system.

# Directions

* On the home page, on the right side (the reference sheet side), choose the language and version, and then the concept.
* The page says the data doesn't exist, but clicking the arrow for the template will provide the template and file for you to use.
* Save that file in `/web/thesauruses/[language]/[version]/[concept].json`.
* Fill out all of the `"code"` sections.
* Add in `"not-implemented": "true"` if the concept doesn't exist in the language. You can then delete the `"code"` line.
* Save, commit your code, push to your fork, then make a PR.
* Watch the checks to make sure they all pass. If they don't, click on "Details" and it should tell you what to modify.

# Claiming This Issue

# Need Help?

# Thanks!

pulgamecanica:

Resolves #426 

## What changed and why?

   I implemented the Strings module for the Ruby 3.x lang 😄 
   I used the official Ruby [documentation](https://ruby-doc.org/core-3.1.2/String.html) as well as [The Well-Grounded Rubyist](https://www.manning.com/books/the-well-grounded-rubyist-third-edition) in case you want to add some references in the future

## Checklist

### DIFF
--- a/web/thesauruses/ruby/3/strings.json
+++ b/web/thesauruses/ruby/3/strings.json
@@ -7,14 +7,14 @@
   },
   "concepts": {
     "is_primitive_or_not": {
-      "comment": "Yes"
+      "code": "Yes"
     },
     "import": {
-      "comment":
+      "code":
         "Built-in included in the default Modules"
     },
     "default_string_byte_encoding": {
-      "comment": "UTF-8"
+      "code": "UTF-8"
     },
     "create_new_string": {
       "code": [
@@ -47,7 +47,7 @@
       ]
     },
     "max_length_of_string": {
-      "comment": "65,535 characters."
+      "code": "65,535 characters."
     },
     "clear_string": {
       "code": [
--- a/web/thesauruses/ruby/3/strings.json
+++ b/web/thesauruses/ruby/3/strings.json
@@ -87,7 +87,10 @@
     "concatenate_two_strings": {
       "name": "Concatenate/join two strings",
       "code": [
-        ""
+        "a = \"String\"",
+        "b = a + \"2\"", 
+        "c = a + b",
+        "puts c # Returns \"StringString2\""
       ]
     },
     "concatenate_many_strings": {
@@ -95,17 +98,14 @@
       "code": [
         "a = \"String\"",
         "b = a + \"2\"", 
-        "c = a + b",
-        "puts c # Returns \"StringString2\""
+        "c = a + \" \" + b",
+        "puts c # Returns \"String String2\""
       ]
     },
     "is_all_alphabetical": {
       "name": "Is string all alphabetical characters?",
       "code": [
-        "a = \"String\"",
-        "b = a + \"2\"", 
-        "c = a + \" \" + b",
-        "puts c # Returns \"String String2\""
+        "!string.match(/\\A[a-zA-Z]*\\z/).nil?"
       ]
     },
     "is_all_numerical": {
--- a/web/thesauruses/ruby/3/strings.json
+++ b/web/thesauruses/ruby/3/strings.json
@@ -1,0 +1,416 @@
+{
+  "meta": {
+    "language": "ruby",
+    "language_name": "Ruby",
+    "structure": "strings",
+    "language_version": "3"
+  },
+  "concepts": {
+    "is_primitive_or_not": {
+      "name": "Is this a built-in type in this language?",
+      "code": [
+        "Yes"
+      ]
+    },
+    "import": {
+      "name": "Import the string class",
+      "code": [
+        "No need?"
+      ]
+    },
+    "default_string_byte_encoding": {
+      "name": "Default byte encoding (ex: ASCII UTF-8, UTF-16, etc.)",
+      "code": [
+        "UTF-8"
+      ]
+    },
+    "create_new_string": {
+      "name": "Create new string",
+      "code": [
+        "1. varName = \"Hello World\"",
+        "2. varName = 'This also works'",
+        "3. anotherVar = %{another way to do it}\n",
+        "1. & 2. Enclosing a set of characters between quotes/double quotes.",
+        "3. Using the %char-style quoting mechanism generates either a single or doubled quoted string.",
+        "You can preceed the % by a character option and a delimiter",
+        "The most common delimiter is {} like the example, but you can use any non alpha-numerical delimiter."
+      ]
+    },
+    "create_multiline_string": {
+      "name": "Create new multi-line string",
+      "code": [
+        "1. text = <<EOM\nLine1\nAnother Line\nNow this is the last line.\nEOM",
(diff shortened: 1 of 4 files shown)


web/thesauruses/ruby/3/strings.json
```diff
@@ -0,0 +1,416 @@
...
+    "is_primitive_or_not": {
+      "name": "Is this a built-in type in this language?",
+      "code": [
+        "Yes"
+      ]
+    },
+    "import": {
+      "name": "Import the string class",
+      "code": [
+        "Built-in included in the default Modules"
```
geekygirlsarah, line 14: Since this isn't actually code but is really a comment, would you mind changing `"code":` to `"comment":` ?

(same for others below)

pulgamecanica:
Hey! I wish I was more careful, I have one mistake.
I misplaced two questions and I forgot to answer one.

The questions misplaced should be one row above, and I should answer the black space that would be left

***

Screen Shot 2022-10-04 at 18 52 14

***

I feel very noob r, wow, srry.

Tbh I feel like, after all, it's good to have PR's instead of committing directly...

pulgamecanica:
Fixed misplaced section and answered the question about alpha strings.

web/thesauruses/ruby/3/strings.json
```diff
@@ -0,0 +1,416 @@
...
+    "default_string_byte_encoding": {
+      "name": "Default byte encoding (ex: ASCII UTF-8, UTF-16, etc.)",
+      "code": [
+        "UTF-8"
+      ]
+    },
+    "create_new_string": {
+      "name": "Create new string",
+      "code": [
+        "1. varName = \"Hello World\"",
```
geekygirlsarah: I also really want to keep "code" lines to be actual compilable code. So `varName = "Hello World"` is compilable, but the "1." at the beginning isn't. (I know you're using it to mark the line of possible options.) While maybe the system should number the lines (an issue I think I'm going to put in), for now, maybe remove the numberings and keep it to just working code.

(also the comments after should go in a "comment" block instead of "code" lines.)

geekygirlsarah (changes requested):
This looks great! A couple of changes to match some best practices and I'll happily merge this in!

web/thesauruses/ruby/3/strings.json
pulgamecanica: Alright! I get it, I'll change it, it makes a lot of sense! :D

web/thesauruses/ruby/3/strings.json
pulgamecanica: Understood! Yes, it was to make very clear the different options, I will change it and try to make it very understandable.

pulgamecanica:
I will read this before going any further...
[https://docs.codethesaur.us/thesaurus/](https://docs.codethesaur.us/thesaurus/)

pulgamecanica:
## Hello, I wanted to ask about this behavior...
### Whenever I replace "code" for "comment", then the application will display "Unknown" above the comment, this is because the application assumes that "code" exists, but I didn't declare it on the JSON scope, I only declared the "comment".
Screen Shot 2022-10-05 at 11 19 48

### There is one way to fix it, which would be to declare the "code" as an empty array on the JSON scope, but this will leave a blank space above the comment.

Screen Shot 2022-10-05 at 11 23 21

***

Perhaps there should be a condition to check if "code" is empty, and if that's the case, only display the comment line, thus removing the blank space for the empty comment.

pulgamecanica:
I also found out why it says, "Unknown".
I think, if you agree with me, it's a nice change. Whenever the concept IS known, but the "code" is None then return None, instead of "unknown".

I already solved it, if you agree with me I would be more than happy to make this change happen.

pulgamecanica:
## My solution would be this:
***
### On the models, on this function:
https://github.com/codethesaurus/codethesaur.us/blob/229b52766c32ed2fbf8edcbe03dcc06460d650e6/web/models.py#L134

### I would change it to something like this:

```
def concept_code(self, concept_key):
        code = self.concept(concept_key).get("code")
        if not code:
            return None
        if isinstance(code, list):
            code = "\n".join(code)
        return code
```

> This way, whenever the "code" in the JSON object is not defined, you return None. (if not code, return None)

***

### On the views on this function :
https://github.com/codethesaurus/codethesaur.us/blob/229b52766c32ed2fbf8edcbe03dcc06460d650e6/web/views.py#L376

### I would change it to something like this:

```
def format_code_for_display(concept_key, lang):
    if lang.concept_unknown(concept_key):
        return "Unknown"
    if lang.concept_code(concept_key) is None:
        return None
    if lang.concept_implemented(concept_key):
        return highlight(
            lang.concept_code(concept_key),
            get_lexer_by_name(lang.key, startinline=True),
            HtmlFormatter()
        )
    return None

```

> So, if the concept is defined but the code is None, then you return None, instead of "Unknown"

pulgamecanica:
This will actually make this line work:

https://github.com/codethesaurus/codethesaur.us/blob/229b52766c32ed2fbf8edcbe03dcc06460d650e6/web/templates/concept_card.html#L5

And the div will not be created, because "code" it's returning None...

pulgamecanica:
#### Hello, I implemented the **" comment"**, and used the **"not-implemented": true** where I thought it was necessary.
#### I also fixed the ordered list.
***

I will open an issue regarding what I described above, I think it's better, that way we can separate these issues.
Best regards,
André

pulgamecanica:
I think I've finished what was left for this String module, I finally deleted the "name" field from the JSON file, implemented the comments; etc as I described before. I would be more than happy to hear your feedback, and I was excited about creating a PR for the "Unkown" issue if you can have a  look at it at some point.
Cheers 😀

geekygirlsarah:
> I feel very noob r, wow, srry.
> 
> Tbh I feel like, after all, it's good to have PR's instead of committing directly...

Don't feel too bad. I've definitely committed stuff that broke the project before, lol!

Also a LOT of people have submitted stuff that's not exactly right. Probably at least half! So don't feel bad. It's why the review process and the GitHub Actions are there. To help validate things before they go in.

geekygirlsarah:
I'm starting to see a bit better why #558 is there. I'll think on that.

Want to wait on the result of #558, or want to just change "comment" back to "code" and I can pull it in now?

pulgamecanica:
I can change it back to code and save this JSON for when it's resolved.

geekygirlsarah:
That works for me!

pulgamecanica:
I believe it's done :D

geekygirlsarah:
This works for me! Thank you for this addition, and your first contribution to CT! 🎉 

If you are participating in [Hacktoberfest](https://hacktoberfest.com/), don't forget to register so that this PR counts towards that.

PR #556, opened 2022-10-04
clarify operand placement with respect to operators in Python and JavaScript

### LINKED ISSUE #512: Clarify operand placement in language definitions of math operators
# Language

Likely all of them

# Wrong Field(s)

Most languages usually use something along the lines of the operator in between two operands (i.e. `a + b`). Not all languages do, notably something like LISP does `(+ a b)`.

There are also prefix and postfix operators that might get confusing if people don't understand what prefix and postfix are. (As in what's the difference between `++` and `++` when we really mean `++a` and `a++`?)

This would be to go through the languages and add in operators to them. Likely most of them are standard infix ones (`a + b`) except prefix/postfix above. For languages that differ, I can help catch it in PR.

This probably means doing something like:
`"addition_operator": "operand1 + operand2"`
(or something else like `a + b` to make it understandable and thorough).

# Additional notes

Check out the documentation at https://docs.codethesaur.us if you need additional help or ask any questions you might have.

Rishav-12:

 

Partially solves #512 

## What changed and why?

   Clarified how the operator is placed in relation to the operands 

## Checklist

   

   #512 is only partially resolved

### DIFF
--- a/web/thesauruses/javascript/ECMAScript 2020/operators.json
+++ b/web/thesauruses/javascript/ECMAScript 2020/operators.json
@@ -7,35 +7,35 @@
   },
   "concepts": {
     "addition": {
-      "code": "+",
+      "code": "operand1 + operand2",
       "name": "Addition operator"
     },
     "addition_assignment": {
-      "code": "+=",
+      "code": "result += operand1",
       "name": "Addition and assignment operator"
     },
     "subtraction": {
-      "code": "-",
+      "code": "operand1 - operand2",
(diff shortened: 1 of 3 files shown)


geekygirlsarah (approved):
Yes! This is great! It clears up the confusion about how the operators and operands work here.

I'll merge it in. Thank you for your first Code Thesaurus contribution!

This counts towards [Hacktoberfest](https://hacktoberfest.com/), so if you're interested in the shirt/swag, don't forget to register!

PR #559, opened 2022-10-05
Fix "Unknow" comment when the concept IS defined but no code provided.

### LINKED ISSUE #558: "Unknown" when code is missing.
# Description

It will remove the "**Unknown**" message which is shown on the reference page if the concept is VALID but it has no "code" text.
I agree, "**Unknown**" should be displayed, but I think only when the concept is not valid... not when there is no "code", because there might be a "comment" that can explain a concept, and NO "code" at all.

Explanatory image:

Screen Shot 2022-10-05 at 11 19 48

> In the example, the default string byte encoding has a comment which explains the concept, but no "code" is provided.
> ```
> "default_string_byte_encoding": {
>     "name": "Default byte encoding (ex: ASCII UTF-8, UTF-16, etc.)",
>      "comment": "UTF-8"
> }
> ```
> You can see on the image, that "Unknown" is displayed above the comment.

# Requirements

Whenever the JSON file has a valid concept, it should consider that it MAY include "code" or "comment" or both at the same time.

***

## On the models, on this function:
https://github.com/codethesaurus/codethesaur.us/blob/229b52766c32ed2fbf8edcbe03dcc06460d650e6/web/models.py#L134

### I would change it to something like this:

```
def concept_code(self, concept_key):
        code = self.concept(concept_key).get("code")
        if not code:
            return None
        if isinstance(code, list):
            code = "\n".join(code)
        return code
```
> Ensuring that code returns None, when it's not present on the JSON file

***

## On the views on this function :
https://github.com/codethesaurus/codethesaur.us/blob/229b52766c32ed2fbf8edcbe03dcc06460d650e6/web/views.py#L376

### I would change it to something like this:

```
def format_code_for_display(concept_key, lang):
    if lang.concept_unknown(concept_key):
        return "Unknown"
    if lang.concept_code(concept_key) is None:
        return None
    if lang.concept_implemented(concept_key):
        return highlight(
            lang.concept_code(concept_key),
            get_lexer_by_name(lang.key, startinline=True),
            HtmlFormatter()
        )
    return None

```

> So, if the concept is defined but the code is None, then you return None, instead of "Unknown"

***

This way, the template will not create the div for the **"code"** section and the condition will make more sense 

https://github.com/codethesaurus/codethesaur.us/blob/229b52766c32ed2fbf8edcbe03dcc06460d650e6/web/templates/concept_card.html#L5

# Additional Notes

In the end, it could look something like this...

Screen Shot 2022-10-05 at 15 40 12

For the given code:

Screen Shot 2022-10-05 at 15 41 18

### The "Unkown" is gone because the concept was well defined even when the "code" is not present (there is "comment")

Cheers, I would love to hear some feedback about this ;)

pulgamecanica:
… present
Hello, I am André, I hope you can consider this PR

Resolves #558 

## What changed and why?

   
I changed the models.py and views.py in order to avoid the display of the message "Unknown" when the concept IS defined but there is no "code" implemented.

### BEFORE
Screen Shot 2022-10-05 at 15 59 54

### AFTER
Screen Shot 2022-10-05 at 15 40 12

Both images are produced by this JSON file:

```
{
  "meta": {
    "language": "ruby",
    "language_name": "Ruby",
    "structure": "strings",
    "language_version": "3"
  },
  "concepts": {
    "is_primitive_or_not": {
      "name": "Is this a built-in type in this language?",
      "comment": "Yes"
    },
    "import": {
      "name": "Import the string class",
      "comment":
        "Built-in included in the default Modules"
    },
    "default_string_byte_encoding": {
      "name": "Default byte encoding (ex: ASCII UTF-8, UTF-16, etc.)",
      "comment": "UTF-8"
    },
    "create_new_string": {
      "name": "Create new string",
      "code": [
        "varName = \"Hello World\"",
        "varName = 'This also works'",
        "anotherVar = %{another way to do it}\n"
      ],
      "comment": "Using the %char-style quoting mechanism generates either a single or doubled quoted string.\nYou can preceed the % by a character option and a delimiter.\nThe most common delimiter is {} like the example, but you can use any non alpha-numerical delimiter."
    },
    "create_multiline_string": {
      "name": "Create new multi-line string",
      "code": [
        "text = <<EOM\nLine1\nAnother Line\nNow this is the last line.\nEOM\n",
        "text = <<-EOM\nLine1\nAnother Line\nNow this is the last line.\n      EOM\n\n"
      ],
      "comment": "\"HERE\" document (heredoc), generates a string, usually a multiline string.\nUses the << operator followed by a delimiter. The string wont end until the delimiter is founded flush left.\nPlacing a '-' after the << operator will disable the flush left requirement."
    },
    "assign_new_string": {
      "name": "Assign string from another string",
      "code": [
        "a = \"String\"",
        "b = a"
      ]
    }
    ...........
```

## Checklist

Cheers 😃


geekygirlsarah:
As I mentioned in #558, I think this is a good thing to think on but still pondering if this is the ideal solution (or at least a part of the solution). I'll think on it more.

geekygirlsarah (changes requested):
I guess this is fine to leave in, however, I'd like the `validatelanginfofiles` command and GitHub Action to reflect this. Would you mind updating that too? It's in `web/management/commands/validatelanginfofiles.py`.

Also the `test_language_get_concept_code` unit test is failing, I think because it requires the concept code to exist and this lets it not exist. So that will need to be fixed.

geekygirlsarah:
I'm going to close this since it hasn't had any work on it since October.

If you wish to pick this back up @pulgamecanica, you're welcome to! You can either reopen this or make a new pull request to work on it.

Thanks for the work you started on so far!

PR #566, opened 2022-10-10
Update issue-auto-unassign.yml

wroblewskialex:

564
 

Resolves #xxxx

## What changed and why?

   

Pointed to a forked version of the issue-auto-unassign bot, which removes references to Ruby For Good

## Checklist

   

- [ ] I have looked at documentation to ensure I made any revisions correctly

### DIFF
--- a/.github/workflows/issue-auto-unassign.yml
+++ b/.github/workflows/issue-auto-unassign.yml
@@ -10,7 +10,7 @@ jobs:
     name: Unassign issues
     steps:
       - name: Unassign issues
-        uses: wroblewskialex/configurable-unassign-issues@main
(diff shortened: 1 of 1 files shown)


geekygirlsarah (changes requested):
Oops, I might have been unclear...

The `uses` will have to stay `codethesaurus/unassign-issues`. The code that is that bot needs to be updated so it doesn't reference RFG.

It looks like from your code (https://github.com/wroblewskialex/configurable-unassign-issues/blob/main/index.js ) you did that. So you should make a PR from `wroblewskialex/configurable-unassign-issues` -> `codethesaurus/unassign-issues` so those changes can get merged in.

(This PR probably should actually close because it has a merge conflict with a change I already put in to point it to CT's version of the bot.)

PR #574, opened 2022-10-15
Queues and Stacks In Python3

### LINKED ISSUE #536: [Python] Add queues and stacks
# Description

Queues and stacks for Python aren't presently on Code Thesaurus. This issue is to add that concept to the system.

# Directions

* On the home page, on the right side (the reference sheet side), choose the language and version, and then the concept.
* The page says the data doesn't exist, but clicking the arrow for the template will provide the template and file for you to use.
* Save that file in `/web/thesauruses/[language]/[version]/queues_stacks.json`.
* Fill out all of the `"code"` sections.
* Add in `"not-implemented": "true"` if the concept doesn't exist in the language. You can then delete the `"code"` line.
* If you want to validate your file, run `python manage.py validatelanginfofiles`
* Save, commit your code, push to your fork, then make a PR.
* Watch the checks to make sure they all pass. If they don't, click on "Details" and it should tell you what to modify.

# Claiming This Issue

# Need Help?

# Thanks!

TrainedPro:

 

Resolves #536

## What changed and why?

Added queues and stack file describing how the Python functionality would work for it. This was an issue which i was assigned to resolve. The file seems to be as per the requirements.

## Checklist

   

I have also changed some names which were incorrect. An example under circular queues is:
`"name": "Data structure that backs the circular queue",` on line 80 where ti was originally simple queue instead of circular queue. This is did not seem to correct it on the website though, when i ran it on my local machine.

### DIFF
--- a/web/thesauruses/python/3/queues_stacks.json
+++ b/web/thesauruses/python/3/queues_stacks.json
@@ -154,7 +154,7 @@
       "double_ended_queue_destroy_statement": {
         "name": "Destroy/delete a queue",
         "code": [
-          "queue.clear()"
+          "del(queue)"
         ]
       },
       "double_ended_queue_data_structure": {
@@ -288,8 +288,11 @@
       "queue_clear_all": {
         "name": "Clear out all queue items",
         "code": [
-          "queue.clear()"
-        ]
+          "dequeue.clear()"
+        ],
+        "comment": "Queues, SimpleQueues, and LifoQueues don't have a clear function."
+      }
+
       },
       "stack_push_item": {
         "name": "Push an item (add to top)",
--- a/web/thesauruses/python/3/queues_stacks.json
+++ b/web/thesauruses/python/3/queues_stacks.json
@@ -241,7 +241,7 @@
       "queue_dequeue_delete_an_item": {
         "name": "Dequeue an item (remove from front, don't return)",
         "code": [
-          "queue.pop()"
+          "queue.pop()",
           "queue.get()",
           "lifoqueue.get()",
           "simplequeue.get()"
@@ -381,4 +381,3 @@
         "not-implemented": "true"
       }
     }
-  }
--- a/web/thesauruses/python/3/queues_stacks.json
+++ b/web/thesauruses/python/3/queues_stacks.json
@@ -291,8 +291,6 @@
           "dequeue.clear()"
         ],
         "comment": "Queues, SimpleQueues, and LifoQueues don't have a clear function."
-      }
-
       },
       "stack_push_item": {
         "name": "Push an item (add to top)",
--- a/web/thesauruses/python/3/queues_stacks.json
+++ b/web/thesauruses/python/3/queues_stacks.json
@@ -379,3 +379,4 @@
         "not-implemented": "true"
       }
     }
+  }
--- a/web/thesauruses/python/3/queues_stacks.json
+++ b/web/thesauruses/python/3/queues_stacks.json
@@ -1,0 +1,388 @@
+{
+    "meta": {
+      "language": "python",
+      "language_name": "Python",
+      "structure": "queues_stacks",
+      "language_version": "3"
+    },
+    "concepts": {
+      "simple_queue_import_statement": {
+        "name": "Import statement to add simple queue",
+        "not-implemented": "true"
+      },
+      "simple_queue_data_type": {
+        "name": "Queue data type name",
+        "code": [
+          "list"
+        ]
+      },
+      "simple_queue_create_statement": {
+        "name": "Create a queue",
+        "code": [
+          "queue = ['One', 'Two', 'Three']"
+        ]
+      },
+      "simple_queue_create_copy_statement": {
+        "name": "Create a queue from existing queue",
+        "code": [
+          "new_queue = queue"
+        ]
+      },
+      "simple_queue_create_copy_from_list_statement": {
+        "name": "Create a queue from a list of items",
+        "code": [
+          "queue = mylist"
+        ]
+      },
+      "simple_queue_destroy_statement": {
+        "name": "Destroy/delete a queue",
+        "code": [
+          "queue.clear()"
+        ]
+      },
+      "simple_queue_data_structure": {
+        "name": "Data structure that backs the simple queue",
+        "code": [
+          "list"
+        ]
+      },
+      "circular_queue_import_statement": {
+        "name": "Import statement to add circular queue",
+        "not-implemented": "true"
+
+      },
+      "circular_queue_data_type": {
+        "name": "Queue data type name",
+        "not-implemented": "true"
+
+      },
+      "circular_queue_create_statement": {
+        "name": "Create a queue",
+        "not-implemented": "true"
+
+      },
+      "circular_queue_create_copy_statement": {
+        "name": "Create a queue from existing queue",
+        "not-implemented": "true"
+
+      },
+      "circular_queue_create_copy_from_list_statement": {
+        "name": "Create a queue from a list of items",
+                "not-implemented": "true"
+
+      },
+      "circular_queue_destroy_statement": {
+        "name": "Destroy/delete a queue",
+                "not-implemented": "true"
+
+      },
+      "circular_queue_data_structure": {
+        "name": "Data structure that backs the circular queue",
+                "not-implemented": "true"
+
+      },
+      "priority_queue_import_statement": {
+        "name": "Import statement to add priority queue",
+        "code": [
+          "from queue import PriorityQueue"
+        ]
+      },
+      "priority_queue_data_type": {
+        "name": "Queue data type name",
+        "code": [
+          "queue.PriorityQueue"
+        ]
+      },
+      "priority_queue_create_statement": {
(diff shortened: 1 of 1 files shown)


web/thesauruses/python/3/queues_stacks.json
```diff
@@ -0,0 +1,388 @@
...
+    },
+    "concepts": {
+      "simple_queue_import_statement": {
+        "name": "Import statement to add simple queue",
+        "not-implemented": "true"
+      },
+      "simple_queue_data_type": {
+        "name": "Queue data type name",
+        "code": [
+          "list"
```
geekygirlsarah: You use a list, which _can_ work for this, but specifically there's [queue.Queue](https://docs.python.org/3/library/queue.html) which seems to be an official queue in Python with fast enqueues and dequeues. It's in the same library as dequeues and PriorityQueues, which you use below. Since lists are described in the `lists.json` file, this should probably be for the built-in queue objects.

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -0,0 +1,388 @@
...
+      "priority_queue_create_copy_statement": {
+        "name": "Create a queue from existing queue",
+        "code": [
+          "new_queue = queue"
+        ]
+      },
+      "priority_queue_create_copy_from_list_statement": {
+        "name": "Create a queue from a list of items",
+        "code": [
+          "from queue import PriorityQueue\nimport itertools\n\nqueue = PriorityQueue()\n\npriorities = [1, 2, 3]\n\nmylist = ['One', 'Two', 'Three']\n\nfor priority, item in list(zip(priorities, mylist)):\n\tqueue.put((priority, item))"
```
geekygirlsarah: This is sort of inventing the functionality for this, which isn't what you should do. This should be "not-implemented" as well.

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -0,0 +1,388 @@
...
+        ]
+      },
+      "priority_queue_destroy_statement": {
+        "name": "Destroy/delete a queue",
+        "code": [
+          "for item in range(queue.qsize()):\n\tqueue.get()"
```
geekygirlsarah: Shouldn't this just be like `del(queue)` ?

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -0,0 +1,388 @@
...
+      },
+      "double_ended_queue_data_structure": {
+        "name": "Data structure that backs the simple queue",
+        "code": [
+          "list"
+        ]
+      },
+      "stack_import_statement": {
+        "name": "Import statement to add stacks",
+        "not-implemented": "true"
```
geekygirlsarah: I think stacks you can do with a [LIFOqueue](https://docs.python.org/3/library/queue.html) which is essentially a queue but last-in-first-out (which is how a stack works).

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -0,0 +1,388 @@
...
+        ]
+      },
+      "iterate_pointer_data_type": {
+        "name": "Data type of a iterator pointer",
+        "not-implemented": "true"
+      },
+      "iterate_create_pointer": {
+        "name": "Create iterator",
+        "code": [
+          "pointer = 0"
```
geekygirlsarah: If iterators/pointers are "not-implemented", then these should be "not-implemented" too. Otherwise they're integers and not iterators.

geekygirlsarah (changes requested):
This is a lot better! You're getting the idea of how it works. I put comments for some suggestions, and if you fix those, I think I could merge this in!

TrainedPro:
All issues stated have been solved along with some others that were similar. Let me know if you would like any further changes.

Thanks!

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -277,13 +273,13 @@
...
+          "mylist = stack"
         ]
       },
       "queue_clear_all": {
         "name": "Clear out all queue items",
         "code": [
-          "queue.clear()"
+          "stack.clear()"
         ]
       },
```
geekygirlsarah: Is this supposed to be `stack_clear_all` or `queue.clear()` ?

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -277,13 +273,13 @@
       "queue_export_to_list": {
         "name": "Export a list of all queue items",
         "code": [
-          "mylist = queue"
+          "mylist = stack"
         ]
       },
```
geekygirlsarah: It looks like you're just copying stack over to mylist, but also this is for a queue. I'm a bit confused by this.

web/thesauruses/python/3/queues_stacks.json
```diff
@@ -162,48 +160,48 @@
...
-        "code": [
-          "mylist = ['One','Two','Three']\nstack = mylist"
-        ]
+        "not-implemented": "true"
       },
       "stack_destroy_statement": {
         "name": "Destroy/delete a stack",
         "code": [
-          "stack.clear()"
+          "del(queue)"
```
geekygirlsarah: Should this be `del(stack)` ?

geekygirlsarah (changes requested):
This is a lot closer! I'm wondering if there's a few copy/paste things that you didn't update as I'm seeing multiple mixed up areas between stacks and queues concepts but using the opposite for the code. There's a couple of other comments I had in there too. Run back over those and make sure they line up and I think this might be good.

web/thesauruses/python/3/queues_stacks.json
TrainedPro: Yes, it seems that I didn't clarify it.

What i did was that where the category was stack, i used the word stack instead of queue, it doesn't matter as it is just a variable name but i just thought it would be more appropriate.

Some place you can see queue used instead of stack even when the name is stack but that was because it was under a queue category, if this is an issue, just let me know as all i really have to do is replace them.

So it shouldn't matter functionally at all. But still, if it matters in terms of presentation, let me know.

web/thesauruses/python/3/queues_stacks.json
TrainedPro: I think this accidentally got replaced when i was replacing the other instances, I'll have this resolved as this doesn't fit in.

Please let me know if you want me to use the "stack" variable name for the stack category or should i use queue with all of them.

geekygirlsarah: It could also be `del(lifoqueue)` since that's what a stack is.

TrainedPro:
Hello!

I think i fixed all the issues that were presented. I now understand the naming convention you used and renamed the variables according to that. I choose to use stack instead of LifoQueue so it is more simplier and more straight forward as the LifoQueue is being imported and it may cause a bit of confusion.

geekygirlsarah (changes requested):
So I'm starting to realize there's some conflicts in this file. Things like:

* dequeue has .append() and .appendLeft() to add to it but Queue/SimpleQueue/LifoQueue has .put()
* dequeue has .pop() and .popleft() but Q/SQ/LQ has .get()
* dequeue has .clear() and .copy() but Q/SQ/LQ doesn't have an equivalent
* dequeue is iterable, Q/SQ/LQ aren't

So parts of this file are correct and parts aren't, and it's entirely dependent on the type of queue you build with. Which probably means there needs to be some comments in there to explain what works with each, or you need more elaborate code blocks. Something like, for example:
```
    "queue_enqueue_an_item": {
        "name": "Enqueue an item (add to end)",
        "code": [
          "deque.append('Four')",
          "queue.put('Four')",
          "lifoqueue.put('Four')",
          "simplequeue.put('Four')"
        ]
      },
```

That way the user can compare all of the objects and see what works. For things like clearing a queue, you might need something like:

```
     "queue_clear_all": {
        "name": "Clear out all queue items",
        "code": [
          "deque.clear()"
        ],
        "comment": "Queues, SimpleQueues, and LifoQueues don't have a clear function."
      },
```

I think it still needs another look over. I appreciate you continuing to work on this and making it better, but it looks like as you've corrected it, a few other confusions have popped up. Do you mind fixing these?

TrainedPro:
No worries, I'll go through it and I'll do these changes.

TrainedPro:
Ok, so i think i implemented all the changes other than the `.copy()` as in Python, that is not too widely used. It is much more common to copy it by assigning it to another variable, which is what i have used.

geekygirlsarah:
> Ok, so i think i implemented all the changes other than the `.copy()` as in Python, that is not too widely used. It is much more common to copy it by assigning it to another variable, which is what i have used.

I think Python shallow copies it. (So like the variable points to the object in memory, and `var2 = var1` would just point them both to the same memory location. It's not a real copy since if you do something to the first one, you've also done it to the second.)

geekygirlsarah (approved):
I think you resolved the things I noticed, and if there's other problems, someone can file an issue with it.

Thanks for your patience and endurance while working on this! I'll merge it in. Congrats on your first Code Thesaurus contribution! 🎉 

Have a great weekend!

PR #469, opened 2022-02-25
Refactor reference and compare views

cafce25:
## What GitHub issue does this PR apply to?

None

## What changed and why?

- Merge the views for language reference & comparision to reduce code duplication across them & their templates and make changing them easier.
  Added benefits are:
  - you can more easily add an additional language on the URL to compare instead of reference.
  - the new code supports comparing more than 2 languages (although this might not display pretty at too many languages).
  - using the same route to display and handle the form data allows us to switch to django `Form`s more easily down the line

- `MetaInfo` handles loading multiple `(language, version)` tuples & throws more descriptive errors instead of `KeyError` and `FileNotFoundError`

- Versions show in page titles & some more links point to the correct files on github (bottom of the page).

- I made sure the old parameters `lang1` and `lang2` still work and did the same for the old routes `/compare` and `/reference` to ensure old links don't break, but since we don't need to differentiate between the both views index now renders them if it gets `concept` & `lang` parameters.

- Version defaults to latest instead of showing an error if you don't provide one
- unify all `*friendly_name` and `*name` variables into `*name`
- unify all `*key` and `*id` variables into `*key`

## Checklist

- [x] I claimed any associated issue(s) and they are not someone else's
- [x] I have looked at documentation to ensure I made any revisions correctly
- [x] I tested my changes locally to ensure they work
- [x] (If editing Django) I have added or edited any appropriate unit tests for my changes

## Any additional comments or things to be aware of while reviewing?
Oof, those are a lot of random changes to the code, but they all sort of naturally evolved from merging the two views and I had a hard time separating them into their own PRs... If you disagree please let me know and I'll try to split this up into smaller chunks!

PR #598, opened 2022-10-17
Merge views

cafce25:
Since we can't reopen #469 

(From the orignal PR: )

None

## What changed and why?

- Merge the views for language reference & comparision to reduce code duplication across them & their templates and make changing them easier.
  Added benefits are:
  - you can more easily add an additional language on the URL to compare instead of reference.
  - the new code supports comparing more than 2 languages (although this might not display pretty at too many languages).
  - using the same route to display and handle the form data allows us to switch to django `Form`s more easily down the line

- `MetaInfo` handles loading multiple `(language, version)` tuples & throws more descriptive errors instead of `KeyError` and `FileNotFoundError`

- Versions show in page titles & some more links point to the correct files on github (bottom of the page).

- I made sure the old parameters `lang1` and `lang2` still work and did the same for the old routes `/compare` and `/reference` to ensure old links don't break, but since we don't need to differentiate between the both views index now renders them if it gets `concept` & `lang` parameters.

- Version defaults to latest instead of showing an error if you don't provide one
- unify all `*friendly_name` and `*name` variables into `*name`
- unify all `*key` and `*id` variables into `*key`

## Checklist

Oof, those are a lot of random changes to the code, but they all sort of naturally evolved from merging the two views and I had a hard time separating them into their own PRs... If you disagree please let me know and I'll try to split this up into smaller chunks!


web/templates/concepts.html
```diff
@@ -4,43 +4,48 @@
...
         <div class="container">
             <div class="row mt-5">
-                <h1 class="col-12">Reference for {{ lang_friendlyname }} for {{ concept_friendly_name | linebreaksbr }}</h1>
+                <h1 class="col-12">{{ title }}</h1>
             </div>
             <div class="row col-12">
-                <p>The tables below show the syntactically highlighted {{ concept_friendly_name }} topic for this
-                    language.</p>
+                <p>The tables below show the syntactically highlighted {{ concept_friendly_name }} topic for the two
+                    languages.</p>
```
geekygirlsarah: "two languages" is fine for comparing but not the reference.

web/templates/concepts.html
```diff
@@ -4,43 +4,48 @@
...
                 <br />
                 <p>If you see any incorrect information, please help us improve! Feel free to either
                     <a href="https://github.com/codethesaurus/codethesaur.us/issues/new/choose" rel="noopener">create
-                        an issue</a> or use the link at the bottom of the language column to correct any information.</p>
+                        an issue</a> or use the links at the bottom of the language columns to correct any information.
+                </p>
```
geekygirlsarah: Similarly there might be one/two links depending on if they do a compare/reference.

geekygirlsarah (changes requested):

geekygirlsarah:
I have a couple of problems with this PR.

It starts off acting like a refactor (via removing deduplication). I'm here for that, let's do it.

Though you also add in extra classes and move code around. The error checking for files existing and all that probably didn't need to move but it's a bit more clear with Exceptions, so I'm ok with that.

But you also rename several variables. Well, ok, also in the sake of clarity I'm ok with this.

At this point the PR has a lot happening and is starting to get confusing and probably shouldn't be so large for something not really discussed previously. Or at least it should have stopped here.

But you also removed the URL paths, which wasn't really discussed either. And I get why you did it, but in turn it does make it more difficult to add in new comparisons of different things if there's no paths anymore. If you're going that route, I'd probably prefer to keep it at `/compare` or `/languages/` or something and it can take 1 or 2 languages and as the longer term vision aspects of CT get added, we can add more paths to account for it. The other thing is I'd probably rather have a URL like `/compare/lang1/lang2/` rather than `/?lang=one&lang=two` as it's a bit more SEO friendly and looks cleaner.

cafce25:
> But you also removed the URL paths, which wasn't really discussed either. And I get why you did it, but in turn it does make it more difficult to add in new comparisons of different things if there's no paths anymore. If you're going that route, I'd probably prefer to keep it at `/compare` or `/languages/` or something and it can take 1 or 2 languages and as the longer term vision aspects of CT get added, we can add more paths to account for it. The other thing is I'd probably rather have a URL like `/compare/lang1/lang2/` rather than `/?lang=one&lang=two` as it's a bit more SEO friendly and looks cleaner.

Definitely with you on paths vs GET parameters.
Should I just change the links to go to the old routes and we worry about changing urls later?
If I do that should I leave in the code that renders concepts on the index page if it receives parameters?

geekygirlsarah:
> Definitely with you on paths vs GET parameters.

Great!

> Should I just change the links to go to the old routes and we worry about changing urls later? 

Yeah, let's not modify the URLs yet. That should probably be its own issue and thinking through how that will work.

> If I do that should I leave in the code that renders concepts on the index page if it receives parameters?

Yeah, that's probably fine, though it'll probably have to get removed when adding in future things that CT is comparing as it query strings on `/` won't necessarily work on other topics. But don't know that anyone is doing this presently, so I'm not sure that it benefits anything by staying in either.

(Don't know if you've seen https://github.com/orgs/codethesaurus/projects?type=classic but I envision a much larger CT, obviously.)

Also I accepted the PR to add the completed generated template into the bottom of the files, so it conflicts with this. I think it should be probably easy to merge the two (though also sorry about causing a merge conflict here!).

geekygirlsarah (approved):
Alright, it's incredibly delayed, but I think this is good after your latest changes. I'll merge it in.

Thanks for the big refactor/merge!

PR #599, opened 2022-10-17
Improve contributors.js to fetch the number in a single request

cafce25:
## What changed and why?

   Improve contributors.js to only doing a single fetch to get the contributor count.

## Checklist

   


geekygirlsarah (approved):
What!? This is bonkers. I really didn't think this would work, but it clearly does. I didn't know they passed the number of pages as a header, which is pretty cool!

Thanks for working on this!

PR #608, opened 2022-10-19
[JS] Added missing data types concept for ECMAScript 2009 and 2023

ulyssear:

Part of issue #609 

## What changed and why?

Added data types concept files from ECMAScript 2021 (which already exists in the current version of codethesaur.us) to ECMAScript 2009 and ECMAScript 2023.

## Checklist

The same thing may be done for the other missing concepts for ECMAScript 2009 and ECMAScript 2023.

### DIFF
--- a/web/thesauruses/javascript/ECMAScript 2009/data_types.json
+++ b/web/thesauruses/javascript/ECMAScript 2009/data_types.json
@@ -1,0 +1,116 @@
+{
+  "meta": {
+    "language": "javascript",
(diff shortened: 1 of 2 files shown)


geekygirlsarah (approved):
This looks good to me! I can merge it in!  Thanks for continuing to add more of the JS versions!

(Also this is having me wonder if there should be a data type for just a number or integer rather than ones of specific bit sizes...)

PR #615, opened 2022-10-21
Add data_types.json entry for Clojure

jeff303:

No issue opened, just working to add more complete Clojure support.

## What changed and why?

Adding data types support for Clojure

## Checklist

   

- [N/A] I claimed any associated issue(s) and they are not someone else's
- [N/A] (If editing Django) I have added or edited any appropriate unit tests for my changes


geekygirlsarah (changes requested):
I see some missing ID fields like `signed_integer_8_bit` and `unsigned_integer_8_bit` for instance. If they're not there, don't delete them but add "not-implemented" to them as the site shows "Unknown" for them if they're missing. 

(Did you _actually_ test your changes? You checked the box on the PR template.)

image

jeff303:
> I see some missing ID fields like `signed_integer_8_bit` and `unsigned_integer_8_bit` for instance. If they're not there, don't delete them but add "not-implemented" to them as the site shows "Unknown" for them if they're missing.

Fixed.

> (Did you _actually_ test your changes? You checked the box on the PR template.)

Yes, but clearly not thoroughly enough.  I basically started from one of the Java entries (which is also missing a few of these entirely), since the type system is essentially identical, so I thought it would be the shortest path.  I confirmed there are no more `Unknown` rows on the data types page as of the latest commit.

geekygirlsarah:
> Yes, but clearly not thoroughly enough. I basically started from one of the Java entries (which is also missing a few of these entirely), since the type system is essentially identical, so I thought it would be the shortest path. I confirmed there are no more `Unknown` rows on the data types page as of the latest commit.

Ahh, that makes sense. Another PR is trying to work on a nicer way of filling in the empty IDs in a bit of an easier way since not all of the language files are all completely filled out due to meta file changes over time.

geekygirlsarah (approved):

PR #616, opened 2022-10-21
Add operators.json entry for Clojure

jeff303:

No issue opened, just working to add more complete Clojure support.

## What changed and why?

Adding operators support for Clojure

## Checklist

   

- [N/A] I claimed any associated issue(s) and they are not someone else's
- [N/A] (If editing Django) I have added or edited any appropriate unit tests for my changes


web/thesauruses/clojure/1/operators.json
```diff
@@ -0,0 +1,163 @@
...
+    },
+    "absolute_value": {
+      "comment": "Using the built-in JDK method.",
+      "code": "(Math/abs -4)",
+      "name": "Absolute value operator"
+    },
+    "percentage": {
+      "code": "%",
+      "name": "Percentage operator",
+      "not-implemented": "true"
```
geekygirlsarah: Is this % or not implemented? Can't have both.

geekygirlsarah (changes requested):

web/thesauruses/clojure/1/operators.json
jeff303: Sorry, in this case I actually did check the rendered website (which was correct) but forgot to remove it from source).  Should be fixed now.

web/thesauruses/clojure/1/operators.json
```diff
@@ -0,0 +1,223 @@
...
+    "structure": "operators"
+  },
+  "concepts": {
+    "addition": {
+      "code": "(+ 1 2 3)",
+      "name": "Addition operator"
+    },
+    "addition_assignment": {
+      "comment": "Clojure doesn't have the traditional notion of assignment as in procedural languages, but it can apply in some situations (ex: atoms and volatile refs).",
+      "code": "(let [x (volatile! 0)]\n  (doseq [addend (range 10)]\n    (vswap! x + addend))\n  @x)",
```
geekygirlsarah: This is basically inventing the solution, and Code Thesaurus is a reference tool for what exists. So therefore it should probably be "not-implemented". (Same for similar assignments below)

web/thesauruses/clojure/1/operators.json
```diff
@@ -0,0 +1,223 @@
...
+      "name": "Increment (add 1) operator"
+    },
+    "decrement": {
+      "comment": "See note about the Addition and assignment operator above. The `dec` function will decrement the value and return the result, not overwrite any \"variable\".",
+      "code": "(dec 10)",
+      "name": "Decrement (subtract 1) operator"
+    },
+    "exponential": {
+      "comment": "No built in function, but easy enough to write.",
+      "code": "(defn exp [x n]\n  (reduce * (repeat n x)))",
```
geekygirlsarah: Same for these, don't write new functions, just report if the language presently can do it. (so "not-implemented")

web/thesauruses/clojure/1/operators.json
```diff
@@ -0,0 +1,223 @@
...
+      "name": "Exponential operator"
+    },
+    "factorial": {
+      "comment": "No built in function, but easy enough to write.",
+      "code": "(defn ! [n]\n  (reduce *' (range 1 (inc n))))",
+      "name": "Factorial operator"
+    },
+    "absolute_value": {
+      "comment": "Using the built-in JDK method.",
+      "code": "(Math/abs -4)",
```
geekygirlsarah, line 86: Importing library functions still counts, so this is fine.

geekygirlsarah (changes requested):
If you can revise for the few spots you wrote similar functionality and just show it's not built-in, I can merge this in.

web/thesauruses/clojure/1/operators.json
```diff
@@ -0,0 +1,213 @@
...
+      "name": "Decrement (subtract 1) operator"
+    },
+    "exponential": {
+      "not-implemented": "true",
+      "name": "Exponential operator"
+    },
+    "factorial": {
+      "comment": "No built in function, but easy enough to write.",
+      "code": "(defn ! [n]\n  (reduce *' (range 1 (inc n))))",
+      "name": "Factorial operator"
```
geekygirlsarah: My first thought is this is a custom-implementation and shouldn't be here. If it's not built-in, it's `not-implemented: true`. 

Second, I always hate using things like "this is easy" because you don't know the experience level of the person using the tool. It's "easy" to someone like me and you, but might not be to someone newer to Clojure or programming.

web/thesauruses/clojure/1/operators.json
```diff
@@ -0,0 +1,213 @@
...
+      "code": "(bit-shift-right 100 2)",
+      "name": "Right shift bitwise operator"
+    },
+    "right_shift_assignment": {
+      "not-implemented": "true",
+      "name": "Right shift assignment operator"
+    },
+    "ternary": {
+      "comment": "Use the `if` function to provide a ternary operator equivalent.",
+      "code": "(if (> 2 1) \"Yes\" \"No\")",
```
geekygirlsarah: This is also inventing a oslution. Should be `not-implemented` too.

geekygirlsarah (dismissed):
Make the two things `not-implemented` and I'll merge it.

geekygirlsarah (changes requested):
Make the two things not-implemented and I'll merge it.

web/thesauruses/clojure/1/operators.json
jeff303: Done.

web/thesauruses/clojure/1/operators.json
jeff303: I actually think in this case it truly is equivalent.  Since every expression returns a value, this is how you would accomplish the same thing that requires the operator in other languages.  See [here](https://reshmeeauckloo.wordpress.com/2015/12/13/clojure-if-cond-case/) for a better explanation than I can give.

Nonetheless I don't feel that strongly about it, so I marked not-implemented.

geekygirlsarah: I think I also misread it because when I look at it now, it clearly works like a ternary. I can either merge this in or you can roll this back.

geekygirlsarah (approved):

geekygirlsarah:
Alright, I'll merge it in since you fixed the not-implemented part but I didn't hear back on the ternary. (We can always revert it back some other time if you want.)

Thanks again!

PR #629, opened 2022-10-22
Get an empty template of a concept from reference and compare page

### LINKED ISSUE #628: [Minor feature] Get the template of a concept from the compare page
# Description

Add a link, like the current link "Want to add or correct information for ____?", which permits the user to get the whole template of a concept.
With this feature, the user can easily update the concept from the template if needed.

# Requirements

* Add a link "Want to get an empty template of this concept?" which generates a template or redirects to a raw file of the template (like the link "Want to add or correct information for ____?" does).

# Additional Notes

(from @geekygirlsarah: This is for a _filled_ out template that contains existing and not existing data.)

ulyssear:
Added link to get an empty template of a concept from reference and compare page.

Resolves #628

## What changed and why?

   * New API endpoints : 
      * reference `/api/{structure}/{lang}/{version}`
      Returns a filled template of a concept for a given language and version.
      ```JSON
      {
         "meta": { 
               "language": ...,
               "language_name": ...,
               "structure": ...,
               "language_version": ...
          },
         "concepts": {...}
      }
      ```
      * compare `/api/{structure}/{lang1}/{version1}/{lang2}/{version2}`
      ```JSON
      {
         "meta": { 
               "language_1": ...,
               "language_version1": ...,
               "language_2": ...,
               "language_version2": ...,
               "structure": ...,
          },
         "concepts1": {...},
         "concepts2": {...},
      }
      ```
      Returns 404 if there is no thesaurus with this combination. 

   * Added a link at the bottom of the pages to get an empty template of a concept. In the case, some entries are missing in current meta files, the contributors will easily get the concept template.

image

image

image
_When javascript is disabled_

## Checklist


cafce25:
Ok here is 3 things about this approach I'm not particularly fond of:
1. With every compare/reference we now ship a template that usually nobody will use
  1a. I think we only really need a template when the concept doesn't really exist yet since else the file is gonna come with the repo anyways. 
3. This is an essential thing that should work even if js is not enabled on the client side
4. we reintroduce raw file handling in the views which we previously pulled out into the models / code generation.

ulyssear:
> Ok, voici 3 choses à propos de cette approche que je n’aime pas particulièrement:
> 
> 1. Avec chaque comparaison / référence, nous expédions maintenant un modèle que généralement personne n’utilisera 1a. Je pense que nous n’avons vraiment besoin d’un modèle que lorsque le concept n’existe pas encore vraiment, sinon le fichier viendra avec le dépôt de toute façon.
> 2. C’est une chose essentielle qui devrait fonctionner même si js n’est pas activé côté client
> 3. Nous réintroduisons la gestion des fichiers bruts dans les vues que nous avons précédemment extraites dans la génération de modèles / code.

Thanks for sharing your point of view @cafce25  !

We can do a fetch to get a JSON of a template. This permits us to not always load the JSON while loading the reference/compare page.

It was useful for me to get missing entries of the function concept for JavaScript. So I share my solution but if there is a better solution, I only have one thing to say: let's go :)

cafce25:
If we can load the template dynamically when js is available but still link to gh when it's not that would be ideal I think. Also rather than doing file handling in the view you should patch `generate_template` imo.

On a side note, why am I writing in french on your end 😂

geekygirlsarah:
> On a side note, why am I writing in french on your end 😂

Je ne sais pas 🤷‍♀️ 😂

geekygirlsarah:
Alright, I've been following along but I've not been at a computer all day to really _see_ what's been happening here. Now that I've had a chance to pull down the code and see what's happening, I'll chime in.

First, I think the main thing I missed isn't that this is the template, it's the _filled in_ template. So if there's things that are missing, they're filled in. This is nice and helpful and probably would help people out.

via @cafce25: 
> Ok here is 3 things about this approach I'm not particularly fond of:
> 
>     1. With every compare/reference we now ship a template that usually nobody will use
>        1a. I think we only really need a template when the concept doesn't really exist yet since else the file is gonna come with the repo anyways.
> 
>     2. This is an essential thing that should work even if js is not enabled on the client side
> 
>     3. we reintroduce raw file handling in the views which we previously pulled out into the models / code generation.

I appreciate these thoughts. The MVP model (which is more of a MVT model for Django) does mean needing to separate things out. Templates should just be templates, and not logic-filled units. Also we should be keeping page loads small, so adding extra data into the pages does slow things down. (I have fast internet in the city, but people in rural areas or without a lot of income have slower speeds or data caps.) Same for JS, which I admit even having disabled on a lot of sites by default because I don't like getting tracked across sites. (CT doesn't track users, but not everyone knows that.)

via @ulyssear:
> We can do a fetch to get a JSON of a template. This permits us to not always load the JSON while loading the reference/compare page.
> 
> It was useful for me to get missing entries of the function concept for JavaScript. So I share my solution but if there is a better solution, I only have one thing to say: let's go :)

I'm not opposed to an API call to get this as we do an API call for the contributors. It just means setting up a view to respond in that way. It should still include the GitHub link.

via @cafce25:
> If we can load the template dynamically when js is available but still link to gh when it's not that would be ideal I think. Also rather than doing file handling in the view you should patch `generate_template` imo.
> 
> On a side note, why am I writing in french on your end 😂

If it helps, I thought about making a Django admin command to fetch any missing IDs from files. This would both be a CLI command as well as a GitHub Action ran on PRs (but not required to pass). It'd be there mostly as informative. I will probably do this anyway, but do like cafce25's idea of patching `generate_template` too. 

Finally, I notice you use an `aria-hidden="true"` in there. I think I recall hearing that sometimes this messes with screen readers and accessibility software. I'd want to check and make sure that's not an issue before merging this in too.

This functionality IS _VERY_ useful, and I want to keep it. I'm just not sure the best place for it yet. Or maybe it should just be in all the places. Either way, I would like to keep CT small, quick, and accessible, as well as well-structured architecturally. The more complexity that keeps getting added, the more it gets harder to manage if it's not done right.

ulyssear:
Hi!
To have another solution to get a complete template for the missing thesaurus, I've created an issue (#635 ) and a PR (#636 ).
I will update this PR with the API call with a new link.

ulyssear:
I replaced the modals with the details tag, I think it will be better from now. 
As for the `aria-hidden` attribute, I'd rather not have too much problem with anything accessibility-related for now. I reserve that for the future, and for future issues :sweat_smile:

web/views.py
```diff
@@ -461,3 +464,45 @@ def concept_reference(concept_id, name, lang):
...
+        lang,
+        version,
+        structure + ".json"
+    )
+
+    try:
+        with open(file_path, "r") as file:
+            response = file.read()
+    except FileNotFoundError:
+        return HttpResponseNotFound()
```
cafce25: I would be much more comfortable with this if file handling stayed in `models.py`

web/views.py
```diff
@@ -461,3 +464,45 @@ def concept_reference(concept_id, name, lang):
...
+    template = generate_language_template(
+        lang,
+        structure,
+        version
+    )
+
+    response = json.loads(response)
+    template = json.loads(template)
+    response = merge(template, response)
+    response = json.dumps(response, indent=2)
```
cafce25: And this should be done in `generate_language_template` imo, that's the whole point of that function

cafce25 (changes requested):

web/urls.py
```diff
@@ -26,4 +26,7 @@
     # /reference/lang1/
     path('reference/', views.reference, name='reference'),
 
+    # /api/{lang}/{version}/{structure_key}
+    path('api/<str:lang>/<str:version>/<str:structure_key>/', views.api, name='api'),
```
cafce25: Thinking about the future maybe the order should be `structure/language/version` instead cause when comparing there will be multiple languages and that way we can keep it the same order across apis.
ie. I'd prefer `<structure>/<lang1>/<v1>/<lang2>/<v2>` over `<lang1>/<v1>/<lang2>/<v2>/<structure>` down the line.
That way appending a language is easier.
What do you think?

ulyssear: You're right, and it makes more sense to make the API URL like other URLs. I'll edit the endpoint.

geekygirlsarah (changes requested):
First, the unit tests are failing.

But when I tested this on the [review app](https://codethesaurus-test-pr-629.herokuapp.com/reference/?concept=functions&lang=ada%3B2012), I got this: 
image

It looks like it's giving a 500 error on Heroku. When I run it locally, I get another 500:

image

This time it's saying a FileNotFound exception. So I think this needs to be debugged a bit more. 

I could maybe help but my day is pretty full so I wouldn't be able to look until tomorrow.

ulyssear:
@geekygirlsarah I think it's fixed now. Can you confirm that?

geekygirlsarah (approved):
Hey, tests work and the review app shows it works too. And I've looked at the code and think it's good. This should hopefully be helpful to anyone trying to edit a pre-existing not-complete file.

I'll merge it in. Thanks for working on this and revising it after my suggestions!

web/views.py
geekygirlsarah: This is now getting added to #635 / #636

Issue #609, opened 2022-10-19
[Javascript] Add missing concepts for ECMAScript 2009 and 2023


ulyssear:
# Language

Javascript (or to be more precise, ECMAScript 2009 and ECMAScript 2023)

# Missing concepts

These concepts have to be added via PRs :

- [x]  Control Structures
   - #613  
- [x] Data Types
   - #608
- [x] Exception Handling
   - #614 
- [x] Functions, Methods, and Subroutines
   - #630 
- [x] Logical and Mathematical/Arithmetic Operators
   - #631 
- [x] Queues and Stacks (neither ECMAScript 2020 nor 2021 don't have this concept)
   - Issue: #541 
   - #647
   - **OBSOLETE**:  PR #563 which adds the concept for ECMAScript 2020 and 2021
- [x] Strings
   - #634 

# Additional Information

To avoid creating X issues about adding concepts for javascript, I decided to create an issue to track the completion of meta files.

~~Since the "queues and stacks" concept is not present in any version, a new issue may be created for this concept.~~

# Disclaimer

I'm already working on some concepts but some help is welcome indeed!

geekygirlsarah:
The "epic" issue for this works for me!

PR #630, opened 2022-10-22
[JS] Updated/Added functions concept

ulyssear:

Part of issue #609 

## What changed and why?

   Added functions concept for ECMAScript 2009 and 2023 and updated the concept for ECMAScript 2020 and 2021.

## Checklist


web/thesauruses/javascript/ECMAScript 2009/functions.json
```diff
@@ -0,0 +1,113 @@
...
+      "code": [
+        "function method({param1}) {}",
+        "function method({param1, param2}) {}"
+      ]
+    },
+    "void_function_variable_parameters": {
+      "name": "Function that does not return a value and function that takes an unknown number of parameters",
+      "code": [
+        "function method() {",
+        "  var params = Array.prototype.slice.call(arguments); // convert arguments to an array",
```
geekygirlsarah: I think this is kind of "inventing" a solution? Typically I've seen this as something like `void functionName (int a, int b, int c...)` and `c` is the array of variable arguments. This looks like you're sort of hacking the function to grab some params somewhere. (also is `arguments` an actual thing? looks like an undefined variable)

web/thesauruses/javascript/ECMAScript 2009/functions.json
```diff
@@ -0,0 +1,113 @@
...
+    "return_value_function_with_parameters": {
+      "name": "Function that returns a value and takes 1 or more defined parameters",
+      "code": [
+        "function method(param1) { return true; }",
+        "function method(param1, param2) { return true; }"
+      ]
+    },
+    "return_value_function_with_keyword_parameters": {
+      "name": "Function that returns a value and takes 0 or more defined keyword parameters",
+      "code": "function method({param1, param2}) { return true; }"
```
geekygirlsarah: Should this be something like `...({param1: "value", param2: "value"})...` ?

web/thesauruses/javascript/ECMAScript 2009/functions.json
```diff
@@ -0,0 +1,113 @@
...
+      "code": [
+        "var method = function() {",
+        "  var params = Array.prototype.slice.call(arguments); // convert arguments to an array",
+        "  // do something with params",
+        "}"
+      ]
+    },
+    "anonymous_function_no_parameters_with_return": {
+      "name": "Anonymous function that takes no parameters and returns a value",
+      "code": "var method = function() { return true; }"
```
geekygirlsarah: I know there's other forms like `() => { statement; }`. Could you include those as well?

web/thesauruses/javascript/ECMAScript 2009/functions.json
```diff
@@ -0,0 +1,113 @@
...
+  "meta": {
+    "language": "javascript",
+    "language_name": "JavaScript",
+    "structure": "functions",
+    "language_version": "ECMAScript 2009"
+  },
+  "concepts": {
+    "void_function_no_parameters": {
+      "name": "Function that does not return a value and takes no parameters",
+      "code": "function method() {}"
```
geekygirlsarah: Typically I see JS functions on multiple lines (like
```
function method() {
   statement;
}
```

While I know JS doesn't care about whitespace, it's typically the "style" of the language. Would you mind adding `\n` characters to separate out lines, or just put separate lines as part of the `code: []` array?

web/thesauruses/javascript/ECMAScript 2020/functions.json
```diff
@@ -1,34 +1,115 @@
...
+      "name": "Function that returns a value and takes an unknown number of parameters",
+      "code": "function method(...params) { return true; }"
     },
     "anonymous_function_no_parameters": {
-      "code": "() => { function_body(); }",
-      "name": "Anonymous function that takes no parameters"
+      "name": "Anonymous function that takes no parameters",
+      "code": [
+        "const method = function() {}",
+        "const method = () => {}"
```
geekygirlsarah: Now that I'm looking at this, you _do_ have the multiple ways of doing it here. Was that new form added in 2020?

geekygirlsarah (changes requested):
I have a few comments and a question... if you can revise and/or tell me why I don't know what I'm talking about, I can merge it in after that!

web/thesauruses/javascript/ECMAScript 2009/functions.json
ulyssear: `arguments` is an object available inside a function.
But, I've used `slice` on `arguments` which is a bad idea according to the french version of MDN: "because it can prevent some JavaScript engine optimizations".
Here is their proposal to avoid using `slice` method:
```js
var args = (arguments.length === 1 ? [arguments[0]] : Array.apply(null, arguments));
```

References : 
- MDN page of `arguments` (english version): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments
- Warning about `slice` with `arguments` (french version): https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Functions/arguments#utilisation_de_la_d%C3%A9composition_avec_les_arguments

web/thesauruses/javascript/ECMAScript 2009/functions.json
ulyssear: > Should this be something like `...({param1: "value", param2: "value"})...` ?

No, but `param1` and `param2` are variables in this scope, so you can do this :
```js
function method({
    param1, 
    param2 = 'world'
}) {
    console.log({param1, param2}); 
    return true; 
}

method({ param1: 'hello' });
```

web/thesauruses/javascript/ECMAScript 2009/functions.json
geekygirlsarah: For this, I'll let it stay since it's sort of a weird thing built into JS despite not having a direct "variable argument function" coorelation.

web/thesauruses/javascript/ECMAScript 2009/functions.json
geekygirlsarah: Keyword parameters are different than what I think this is. So it'd be you can name your parameters and then supply them in any order, and I think you're doing default parameters.

So I'm guessing this is actually "not-implemented"

geekygirlsarah: I thought this didn't match but I guess it does. Huh!

ulyssear: I tried something today that works too which permits to store attributes with a key including "-" (for example) in variables:
```js
function method({
    'long-name-1': param1 = 42,
    'long-name-2': param2 = 'hello'
  }) {
  console.log({
    param1,
    param2
  });
}
```

geekygirlsarah (approved):
Alright, I think it looks good now. I'll merge it in!

PR #634, opened 2022-10-23
[JS] Added/Updated strings concept

ulyssear:

Part of issue #609 

## What changed and why?

   Added strings concept for ECMAScript 2009 and 2023, and updated for ECMAScript 2020 and 2021.

## Checklist


web/thesauruses/javascript/ECMAScript 2009/strings.json
```diff
@@ -0,0 +1,344 @@
...
+      "language_version": "ECMAScript 2009"
+    },
+    "concepts": {
+      "is_primitive_or_not": {
+        "name": "Is this a built-in type in this language?",
+        "code": "true"
+      },
+      "import": {
+        "name": "Import the string class",
+        "code": "String"
```
geekygirlsarah: You don't have to import the String class in JS, right? I'm thinking this is not-implemented.

web/thesauruses/javascript/ECMAScript 2009/strings.json
```diff
@@ -0,0 +1,344 @@
...
+        "not-implemented": true,
+        "name": "Destroy string"
+      },
+      "length_of_string": {
+        "name": "Length of string",
+        "code": "str.length"
+      },
+      "max_length_of_string": {
+        "name": "Maximum length of string",
+        "code": "str.length = 5;"
```
geekygirlsarah: This is for what the maximum length of a string might be. (like if it was 1024 bytes maximum or something.) I think this is not-implemented since I think JS doesn't have a physical limit to it outside of RAM limits.

web/thesauruses/javascript/ECMAScript 2009/strings.json
```diff
@@ -0,0 +1,344 @@
...
+      },
+      "clear_string": {
+        "name": "Clear/empty string",
+        "code": [
+            "str = \"\";",
+            "str.length = 0;"
```
geekygirlsarah: You can't set a length that way. (I was curious so I tried it. It lets me set it to 0, but then the string is still the same, so it's not actually cleared. I think only the first example works.

image

web/thesauruses/javascript/ECMAScript 2009/strings.json
```diff
@@ -0,0 +1,344 @@
...
+      "concatenate_many_strings": {
+        "name": "Concatenate/join many strings",
+        "code": [
+            "str + str2 + str3 + str4",
+            "str.concat(str2, str3, str4)"
+        ]
+      },
+      "is_all_alphabetical": {
+        "name": "Is string all alphabetical characters?",
+        "code": "(/^[a-zA-Z]+$/g).test(str)"
```
geekygirlsarah: If you're writing a regex for this, you're "inventing" a solution. I know there's functions like `isInteger()` for some things but not sure if there is for all of these `is_all_` things here. But don't use regexes, just only use if there's built-in ways to determine this in the language.

geekygirlsarah: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#static_properties  is an example for Number objects, you could look at Strings and such too

web/thesauruses/javascript/ECMAScript 2009/strings.json
```diff
@@ -0,0 +1,344 @@
...
+        "code": [
+            "str.indexOf(substring)",
+            "str.search(substring)"
+        ]
+      },
+      "find_start_index_of_additional_substring": {
+        "name": "Find index of an additional substring (or starting at another index)",
+        "code": [
+            "str.indexOf(substring, index)",
+            "str.search(substring)"
```
geekygirlsarah: `search()` only finds the first instance of it. You could write a regex to find others, but that's really the `match()` function. I think just keep the `indexOf()`

(Similar for other `search()` below.)

web/thesauruses/javascript/ECMAScript 2023/strings.json
```diff
@@ -0,0 +1,352 @@
...
+            "str.split(substring).length - 1",
+            "str.match(/substring/g).length"
+        ]
+      },
+      "get_leftmost_characters": {
+        "name": "Get a specified number of characters from the left",
+        "code": [
+            "str.slice(start, end)",
+            "str.substring(start, end)",
+            "str.substr(start, length)"
```
geekygirlsarah: According to MDN, they say substr() is deprecated and not to use it anymore. (I'm not sure when it got deprecated, so it may or may not exist in 2009/2020/2021/2023.) It should probably be removed in 2021/2023 at least.

web/thesauruses/javascript/ECMAScript 2023/strings.json
```diff
@@ -0,0 +1,352 @@
...
+        "name": "Replace all substring matches with another string",
+        "code": [
+            "str.replace(/substring/g, another)",
+            "str.split(substring).join(another)"
+        ]
+      },
+      "split_at_index": {
+        "name": "Split string into a list of strings at a given index",
+        "code": "str.split(separator, limit)",
+        "comment": ""
```
geekygirlsarah: Did you mean to leave an empty comment?

web/thesauruses/javascript/ECMAScript 2023/strings.json
```diff
@@ -0,0 +1,352 @@
...
+        "name": "Decode string from Base64 format",
+        "code": "atob(str)"
+      },
+      "format_string_function": {
+        "name": "Function to format a string",
+        "code": [
+          "'' + str",
+          "String(str)"
+        ]
+      },
```
geekygirlsarah: This is for like functions to format strings in various ways. You are just coercing things into strings, which is not the same. It looks like the various `*.format()` functions you ahve below are what you use, but since they all vary and it's not one function to do them all, I think this is `not-implemented`, right?

geekygirlsarah (changes requested):
A few comments in here, but I think this is close to a merge. Let me know if I'm missing anything.

web/thesauruses/javascript/ECMAScript 2009/strings.json
ulyssear: It's right, `String` is a built-in JavaScript class. 
You can do something like this :
```js
const string4 = new String("A String object");
```
But I can still add `not-implemented` instead.

References : 
- MDN page about String (with the code above): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#creating_strings

web/thesauruses/javascript/ECMAScript 2009/strings.json
ulyssear: OK ! I'll change that.

web/thesauruses/javascript/ECMAScript 2009/strings.json
ulyssear: My bad, I wrote that too fast, I'll edit that too

web/thesauruses/javascript/ECMAScript 2023/strings.json
ulyssear: I'll remove that (if I have no comment to write there)

web/thesauruses/javascript/ECMAScript 2023/strings.json
ulyssear: Yes, I'll add `not-implemented`!

geekygirlsarah (approved):
I think this looks right but I was having trouble testing it. (I blame Heroku.) I'll go ahead and merge it since I don't see issues and the tests/validators pass.

Thanks again!

PR #652, opened 2022-11-01
Fix request instead of referer

cafce25:

 

None

## What changed and why?

Use referer instead of request
   

   n/a

## Checklist

   


cafce25:
Maybe we can recover the referrer data from before, but we would have to look into the concrete data written to the DB before this change. Anyway once we apply this we should clean up the DB.

geekygirlsarah (approved):
Tested it locally and it works!  I feel like this lends itself to some unit tests to help verify this and the other DB writing things, but I think that's another ticket.

I can merge it in.  As for old data, I'll just have to change it (since I'm presently the only one with access). I guess I'll have to look through it to see if anything besides paths are rescuable.

Issue #689, opened 2023-10-24
BUG: Go's queues/stacks fails


geekygirlsarah:
Error report:

```
Internal Server Error: /reference/

AttributeError at /reference/
'MetaInfo' object has no attribute 'language_friendly_name'

Request Method: GET
Request URL: https://codethesaur.us/reference/?concept=queues_stacks&lang=go%3B1
Django Version: 4.2.3
Python Executable: /app/.heroku/python/bin/python
Python Version: 3.11.4
Python Path: ['/app/.heroku/python/bin', '/app', '/app/.heroku/python/lib/python311.zip', '/app/.heroku/python/lib/python3.11', '/app/.heroku/python/lib/python3.11/lib-dynload', '/app/.heroku/python/lib/python3.11/site-packages']
Server time: Tue, 24 Oct 2023 15:22:05 +0000
Installed Applications:
['web.apps.WebConfig',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'markdownify.apps.MarkdownifyConfig']
Installed Middleware:
('django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.common.BrokenLinkEmailsMiddleware')


Traceback (most recent call last):
  File "/app/web/models.py", line 278, in load_languages
    language.load_concepts(meta_structure.key, version)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/web/models.py", line 91, in load_concepts
    with open(file_path, 'r', encoding='UTF-8') as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

During handling of the above exception ([Errno 2] No such file or directory: 'web/thesauruses/go/1/queues_stacks.json'), another exception occurred:
[...]

PR #690, opened 2023-10-24
Fix AttributeError in MetaInfo class

JureBl:

BUG: Go's queues/stacks fails #689

## What changed and why?

   Fix AttributeError in MetaInfo class

Replace incorrect 'self.language_friendly_name(language_key)' with 'self.language_name(language_key)' in the load_languages method of the MetaInfo class. 

## Checklist

   

### DIFF
--- a/web/models.py
+++ b/web/models.py
@@ -242,6 +242,7 @@ class MetaInfo:
             meta_info_json = json.load(meta_file)
         self.structures = meta_info_json["structures"]
         self.languages = meta_info_json["languages"]
+        
 
     def language_name(self, language_key):
         """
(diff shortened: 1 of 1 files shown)


geekygirlsarah:
I made the review app https://codethesaurus-test-pr-690.herokuapp.com . Go's queues/stacks aren't listed. Are there other good ways to test this?

Also could unit tests be added to this somehow to verify it's proper working order?

JureBl:
```
    def test_language_bad_key_and_lang_exists(self):
        """test Language behaviour with bad language key"""
        language = self.dummy_language

        self.assertEqual(bool(language), False)
        self.assertRaises(FileNotFoundError,
                          language.load_concepts,  "notastructure", "notaversion")
```

I think that is already testing a language that is not listed.  Or for what case do you need a Unit-test? Also I am not very familiar with Unit-tests :)

geekygirlsarah:
Yeah, but the test apparently didn't check something because we were getting 500 server errors. If you're not familiar with them, maybe I'll see if I can add a test soon.

I'll go ahead and merge this in.

PR #659, opened 2023-01-19
Akkshay tandon  adding c language

akkshayTandon:
## What GitHub issue does this PR apply to?

 

none

## What changed and why?

 added strings for C language

## Checklist

   

- [x] I claimed any associated issue(s) and they are not someone else's
- [x] I have looked at documentation to ensure I made any revisions correctly
- [ ] I tested my changes locally to ensure they work
- [ ] (If editing Django) I have added or edited any appropriate unit tests for my changes

PR #714, opened 2023-11-30
Akkshay tandon  adding c language

akkshayTandon:
None, (in reference to the closed PR #659 )

 

## What changed and why?
- updated Strings for C language

## Checklist

   

- [ ] I have looked at documentation to ensure I made any revisions correctly


web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+        "char s[] = \"hello\";",
+        "strcpy(s,\"\");"
+      ],
+      "name": "Clear/empty string"
+    },
+    "is_empty": {
+      "code": [
+        "int result = strcmp(s,\"\");",
+        "printf(\"%d\",result);"
+      ],
```
geekygirlsarah: This works but it's not a built-in function to do this, you're using a different function to emulate it. I think this should be `not-implemented`

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "name": "Is string empty?"
+    },
+    "concatenate_two_strings": {
+      "code": [
+        "char str1[] = \"hello\";",
+        "char str2[] = \" world\"",
+        "strcat(str1,str2);",
+        "puts(str1);"
+      ],
```
geekygirlsarah: _I_ know this puts `str2` into `str1` but it might be helpful to add a `comment` to this to make that more clear

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "name": "Return a substring from a string based on starting index and size of substring"
+    },
+    "convert_to_uppercase": {
+      "code": [
+        "#include<stdio.h>",
+        "#include<string.h>",
+        "char str[100];",
+        "gets(str);",
+        "printf(\"%s\",strupr(str);"
+      ],
```
geekygirlsarah: You can probably make this simpler and/or more direct by just using a pre-coded string. like:
```
#include<stdio.h>
#include<string.h>
char str[] = "Hello, world!";
printf(\"%s\",strupr(str);
```

Also at first I thought this was `isupper()` and you were wrong, but turns out we're both right! Maybe adding that in as a second option? (same for `islower()`)

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "not-implemented": "true",
+      "name": "Convert string to title case"
+    },
+    "capitalize_string": {
+      "not-implemented": "true",
+      "name": "Capitalize first letter of a string"
+    },
+    "remove_whitespace": {
+      "not-implemented": "true",
+      "comment":"isspace() is available for custom implementation",
```
geekygirlsarah: I think you can remove this. The idea of a thesaurus wouldn't suggest how to make your own.

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "not-implemented": "true",
+      "name": "Paramater used in format function if they're named variables"
+    },
+    "format_as_integer": {
+      "code":[
+        "printf(\"%d\", age); // %i can be also used",
+        "printf(\"%u\", age); // %u for unsigned integer format"
+      ],
+      "name": "Format parameter as an integer"
+    },
```
geekygirlsarah: I guess this is a bit vague. `%d` is for signed integers, `%i` for unsigned, and perhaps these should be two different things. I can add this into the meta language file.

but for now, maybe clear it up by saying which type of int `%d/%i` are and the differences.

geekygirlsarah (changes requested):
I thought I had reviewed this, but I think I reviewed your last PR and then you were updating this one for a while that I never finished a review for it. Sorry I dropped the ball on that!

Here's a couple of comments to clear up, then I can merge it in!

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+    "structure": "strings"
+  },
+  "concepts": {
+    "is_primitive_or_not": {
+      "code": "no",
+      "name": "Is this a built-in type in this language?"
+    },
+    "import": {
+      "code": "#include<string.h>",
+      "name": "Import the string class"
```
cafce25: There are no classes in C so the `"name"` can't be right. I'd write something along the lines of "Include string related function and type definitons." maybe without the "related".

web/thesauruses/c/17/strings.json
cafce25: You would think it does, but since both `str1` and `str2` only have the exact space the string literals they are initialized with available, this invokes UB as it tries to write past the end of `str1`. C is hard 🙃

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "code": [
+        "int result = strcmp(s,\"\");",
+        "printf(\"%d\",result);"
+      ],
+      "name": "Is string empty?"
+    },
+    "concatenate_two_strings": {
+      "code": [
+        "char str1[] = \"hello\";",
+        "char str2[] = \" world\"",
```
cafce25: missing a `;`

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "name": "Create new string"
+    },
+    "create_multiline_string": {
+      "not-implemented": "true",
+      "name": "Create new multi-line string"
+    },
+    "assign_new_string": {
+      "code": [
+        "char s[] = \"hello\";",
+        "char st[] = s;",
```
cafce25:
```
<source>:5:17: error: invalid initializer
    5 |     char st[] = s;
```

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "name": "Assign string from another string"
+    },
+    "destroy_string": {
+      "not-implemented": "true",
+      "name": "Destroy string"
+    },
+    "length_of_string": {
+      "code": [
+        "chat str[] = \"hello\";",
+        "printf(\"%s\",strlen(s));"
```
cafce25: `%s` is the wrong format specifier for the `size_t` returned by `strlen`

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+    "convert_to_lowercase": {
+      "code": [
+        "#include<stdio.h>",
+        "#include<string.h>",
+        "char str[100];",
+        "gets(str);",
+        "printf(\"%s\",strlwr(str);"
+      ],
+      "name": "Convert string to all lowercase"
+    },
```
cafce25: [`strupr` and `strlwr` are broken, non standard functions](https://stackoverflow.com/questions/26327812/are-strupr-and-strlwr-in-string-h-part-of-the-ansi-standard)

web/thesauruses/c/17/strings.json
```diff
@@ -0,0 +1,302 @@
...
+      "not-implemented": "true",
+      "name": "Maximum length of string"
+    },
+    "clear_string": {
+      "code": [
+        "char s[] = \"hello\";",
+        "strcpy(s,\"\");"
+      ],
+      "name": "Clear/empty string"
+    },
```
cafce25: I think the same concerns as Sarah has for `is_empty` apply.

cafce25 (changes requested):

web/thesauruses/c/17/strings.json
cafce25:
```suggestion
      "code": [
        "char str1[12] = \"hello\";",
        "char str2[] = \" world\";",
        "strcat(str1,str2);",
        "puts(str1);"
      ],
```

web/thesauruses/c/17/strings.json
akkshayTandon: I am a little confused here. I read a few articles on internet regarding different types of format specifiers in C. Some suggest %i as signed some as unsigned and some are neutral. But %d for signed and %u for unsigned are consistent across all.
So, I think only these two should be there to clear confusion. 

what's your say?

cafce25:
I'm not sure why you've requested a review, you haven't fixed any of the points I brought up, at least not in a way that is visible here. Did you maybe forget to commit/push your changes so they're not visible here?

akkshayTandon:
@cafce25 I have pushed the changes

geekygirlsarah (approved):
I think the issues I and cafce25 had look resolved to me. I'll merge it in.

Thanks for working on this for so long, and going back and forth with us on revisions!

PR #777, opened 2025-09-05
Fix: Show friendly 500.html when database is unavailable

### LINKED ISSUE #776: [Bug] Internal Server Error if database connection isn't available
# Expected Behavior

If the database loses connection (like during upgrades), it just shows a white screen with "Internal Connection Error" (or something along those lines).

# Actual Behavior

A nicer page along the lines of "The page is temporarily down, we'll be back shortly." 

1. Locally, remove the database or in the settings file point it to somewhere that doesn't exist
1. Visit the home page

ashan262:

 

Resolves #776 

## What changed and why?

This PR improves error handling when the database is unavailable.  
Previously, users would see the default Django error output, which exposed technical details and was not user-friendly.  
### Changes made:
1. **Created `DatabaseDownMiddleware`** in `web/middleware.py`  
   - Intercepts database errors and ensures users are redirected to a custom error page.  

2. **Registered the middleware** in `settings.py`  
   - Added:  
     ```python
     'web.middleware.DatabaseDownMiddleware'
     ```  

3. **Updated `error500.html` page** in the templates directory  
   - Shows a clear and user-friendly message:  
     *“The site is temporarily down, please try again later.”*  

This ensures users see a consistent and friendly downtime page instead of a raw Django traceback.  

### Add a file web/middleware.py for catching database operational error:
image

### Add the DatabasedownMiddleware from middleware.py to MIDDLEWARE in setting.py
image

### Message change in error500.html:
image

### Finally, this is how It will look like: 
image 

## Checklist

   

- [ ] For language files, I have validated the edited files are valid JSON and data shows up correctly
- [ ] For website code edits, I have added or edited any appropriate unit tests for my changes

To test this change locally, please ensure the following settings are applied in `settings.py`:

```python
DEBUG = False
SECURE_SSL_REDIRECT = False

```
- **DEBUG = False** → Required because custom error pages (error500.html) only render when Django is not in debug mode. With DEBUG=True, Django displays the default debug traceback instead.

- **SECURE_SSL_REDIRECT = False **→ Disabled so the Django development server doesn’t attempt to force HTTPS (it only supports HTTP locally).

With these settings, when the database connection is unavailable, the updated error500.html page will be shown to users instead of the default Django error response.

### DIFF
--- a/codethesaurus/settings.py
+++ b/codethesaurus/settings.py
@@ -49,6 +49,7 @@ INSTALLED_APPS = [
 MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'web.middleware.DatabaseDownMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
--- a/web/middleware.py
+++ b/web/middleware.py
@@ -1,0 +1,18 @@
+# web/middleware.py
+from django.db.utils import OperationalError
+from django.shortcuts import render
+from django.http import HttpResponseServerError
+
+class DatabaseDownMiddleware:
+    """
+    Catch database OperationalError and show custom 500 error page.
+    """
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        try:
+            return self.get_response(request)
+        except OperationalError:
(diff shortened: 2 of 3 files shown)


geekygirlsarah (approved):
Huh, I hadn't thought about how to solve this yet but hadn't thought about using middleware. Makes sense to me!

Thanks for tackling this! Have a great weekend!

ashan262:
@geekygirlsarah 
Thanks! Gave the server a little guidance—middleware to the rescue.
Hope your weekend is less “error 500” and more chill!

